PR #4, opened 2015-12-03
TextInput submit on return

burgalon:


FaridSafi:
Thank you Alon! I will merge and set it as an option

PR #13, opened 2015-12-17
[GiftedMessenger] Fix issue with RN 0.16.0

wenkesj:
Multiple events propagating, ensure only one goes through to not throw exception.


wenkesj:
@FaridSafi I had an issue with the current branch of RN, would look into propagation issues. I appreciate this module a lot and think it has HUGE potential. It serves well in my groups application.

PR #14, opened 2015-12-21
visual bug fix in textInputContainer on iPhone 6 Plus where the pixel…

talkol:
visual bug fix in textInputContainer on iPhone 6 Plus where the pixel ratio is x3 and not x2 so 0.5 should be 0.333


FaridSafi:
Great! Thank you @talkol, I will merge it tomorrow

FaridSafi:
The fix has been published, thanks!

PR #23, opened 2016-01-16
Changed bubbles so they size to text instead of static width.

slavik0329:
This makes the chat bubbles look more like native messanger


slavik0329:
For reference: 

Screenshot

slavik0329:
I messed up these two pull requests again. Maybe 3rd time will be the charm!

PR #25, opened 2016-01-16
Style changes for bubbles

slavik0329:
Bubbles now look like native IOS messaging bubbles with variable with
Username positioned more appealingly

For reference: 

Screenshot


FaridSafi:
Thank you @slavik0329, this is great!

PR #30, opened 2016-01-19
Added onChangeText prop

slavik0329:
I needed this to implement a "Typing..." feedback for a chat. There was no way to access the textinput changes without implementing this patch. I figure it can be valuable for anyone else as well.


FaridSafi:
Hello @slavik0329, Thank you for this PR.
I merged manually and I published it in version 0.0.14

PR #36, opened 2016-01-23
fix rendering img and status of My msgs

kesha-antonov:


FaridSafi:
@kesha-antonov Thank you. I guess this bug was introduced after 0.0.11

kesha-antonov:
Thanks! Yeah seems like that. That was a big good PR.

PR #40, opened 2016-02-03
Scroll to bottom on messages prop change + Style fix

slavik0329:
When messages prop is changed, we should scroll to the bottom to get latest messages into view
Name style fix

### DIFF
--- a/GiftedMessenger.js
+++ b/GiftedMessenger.js
@@ -356,6 +356,8 @@ var GiftedMessenger = React.createClass({
 
     this.setState({
       dataSource: this.state.dataSource.cloneWithRows(this._data, this._rowIds),
+    }, ()=>{
+      this.scrollToBottom();
     });
 
     return rowID;
--- a/Message.js
+++ b/Message.js
@@ -10,7 +10,7 @@ var styles = StyleSheet.create({
   name: {
     color: '#aaaaaa',
     fontSize: 12,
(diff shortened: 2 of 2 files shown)


niftylettuce:
@slavik0329 Is this related to https://github.com/FaridSafi/react-native-gifted-messenger/issues/35?

slavik0329:
@niftylettuce yes, I think this should solve that problem

niftylettuce:
@slavik0329 Unfortunately it didn't :frowning:, I just installed your fork with this commit and it hasn't resolved it.

slavik0329:
@niftylettuce ah, nevermind. I thought you were having the same issue where the messages did load, but were just hidden from view

niftylettuce:
Figured it out, I had to upgrade to RN 0.18+ because of `onLayout`

ckurban:
I have similar issue, in iso (havent tried in android yet) after loading earlier messages from remote it does not scroll to bottom, is what you see guys ?

koppelaar:
A related issue is that when there are only a few messages in the listview (so the messages do NOT fill up the screen) and the keyboard is shown, it is customary for chat apps to scroll to the bottom. Users are typically replying to the last message and this scroll action ensures that the last message is still visible. This scroll action usually happens at the same time the keyboard is becoming visible.

PR #46, opened 2016-02-07
Update README.md

niftylettuce:
https://github.com/FaridSafi/react-native-gifted-messenger/blob/master/GiftedMessenger.js#L452


FaridSafi:
Thank you

Issue #2, opened 2015-11-30
Split this component in more components


sibeliusseraphini:
In renderRow is better to have another component to render it, like <Message>

So, in next implementations we could have different types of Messages, like a message that has a photo, or a Gif or Location.

Or a <Message> component could have a List of <MessageParts>, in this design we could have in the same Message more than one type of message, like text + gif + photo

olivierlesnicki:
+1

FaridSafi:
What do you think about adding a view param to the messages :
message = {
  view: <View><Image /></View>
  // ...
}
If a view is specified, it will be added inside the bubble.
The view can contains all kinds of custom components - image/location/gif/video/etc.

An "Import media" button will also be needed close to the TextInput to trigger an PopUp/ActionSheet

olivierlesnicki:
I don't think it's a wise idea to put the `view` element inside the `message` object. Instead we should expose a: `renderMessage(message) {}`

FaridSafi:
what about getInitialMessages ?

sibeliusseraphini:
we should have a **type** inside each message, and the type will map it to the right component to render it

for instance
type: 'text' -> <MessageText />
type: 'image' -> <MessageImage />
type: 'gif' -> <MessageGif />
type: 'location' -> <MessageLocation />

one problem is how to make easy to add more message types

FaridSafi:
I'm thinking about a solution similar to Media Attachment of https://github.com/jessesquires/JSQMessagesViewController

The media component will have to be added to the message object
Some default media components will be provided
The media can be a custom component following some rules

sibeliusseraphini:
it could work, I will try to code something in the next two weeks

FaridSafi:
fixed by 0.0.11

FaridSafi:
Fixed by 0.0.18.
Thanks @tommoor 

Example:
```js
message = {
  view: <View><Image /></View>
  // ...
}
```

PR #49, opened 2016-02-12
Allow passing message component as prop

tommoor:
Simply pass a `view` property with the rowData to render a custom component, defaults to Bubble. This is a simple solution that provides a lot of flexibility.

should close out #2


FaridSafi:
version 0.0.18 published
Thank you man

PR #53, opened 2016-02-16
scrollWithoutAnimationTo(vertical, horizontal)

Alchemist85K:
scrollWithoutAnimationTo(vertical, horizontal)

### DIFF
--- a/GiftedMessenger.js
+++ b/GiftedMessenger.js
@@ -188,7 +188,7 @@ var GiftedMessenger = React.createClass({
           forceRenderImage={this.props.forceRenderImage}
           onImagePress={this.props.onImagePress}
           renderCustomText={this.props.renderCustomText}
-          
+
           styles={this.styles}
         />
       </View>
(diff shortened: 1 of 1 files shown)


FaridSafi:
Thank you, do you know if it's related to some react-native core breaking change?

FaridSafi:
It seems that will be different in 0.20
https://github.com/facebook/react-native/commit/6941c4e027abb582d7cef0e7c3b9b7ebf51bc070

FaridSafi:
before I publish a new version, do you confirm https://github.com/FaridSafi/react-native-gifted-messenger/blob/master/GiftedMessenger.js#L257 doesn't need to be changed as well ?

Alchemist85K:
No, I didn't check it.
But I think it would be okay because the first parameter of `scrollTo()` function is y value.
And Would it be better to replace `scrollWithoutAnimationTo()` with `scrollTo({x, y, animated: false})`?

FaridSafi:
I guess ```js scrollTo({x, y, animated: false}) ``` is not working yet with 0.19

Alchemist85K:
Oh, I got it. Your project helps me a lot. Thank you.

PR #59, opened 2016-02-23
Add loadMessagesLater flag to props

alizbazar:
This makes it possible to render the messenger while messages are still being fetched.

Before if no messages were supplied, `allLoaded` was set to `true` assuming there are no earlier messages to be loaded. Now by setting `loadMessagesLater` flag the messenger won't make that assumption.


FaridSafi:
Thank you @alizbazar

PR #66, opened 2016-02-28
Minor usability fixes to the UX

alizbazar:
- Enable sending message with return key (this should be default behavior)
- Scroll to bottom after send (the functionality is there, it's just not used atm.)
- Remove styles.imagePosition from avatar pic's TouchableHighlight (currently styles.imagePosition is used both for the Image element **and** the TouchableHighlight, which is probably a mistake)


FaridSafi:
Thank you @alizbazar 
These fixes were needed

PR #71, opened 2016-03-06
Expose full rowData to renderCustomText and add dynamic hide/show text input

### LINKED ISSUE #65: renderCustomText returns incorrect arguments
When I define ``renderCustomText`` my callback is passed ``(rowData, position)`` rather than ``(rowData, rowID)`` which means I can't get the message object inside my callback.

### LINKED ISSUE #72: Expose full rowData object to renderCustomText
Needed to create specific rendered components based on other rowData attributes.

### LINKED ISSUE #73: Make text input hide/show dynamic
When hiding/showing test input change animated height to be dynamic

bpeters:
Resolves #72 
Resolves #73
Resolves #65 

For those who want to render content based on specific rowData attributes, I exposed the full rowData to the renderCustomText function.

Text input hide/show needs to be dynamic, so I changed the animated height to make it work.

Added autoScroll props as a new message array comes in, it'll auto scroll to bottom.

### DIFF
--- a/Bubble.js
+++ b/Bubble.js
@@ -45,7 +45,7 @@ export default class Bubble extends React.Component {
   renderText(text = "", position) {
 
     if (this.props.renderCustomText) {
-      return this.props.renderCustomText(text, position);
+      return this.props.renderCustomText(this.props);
     }
     return (
       <Text style={[styles.text, (position === 'left' ? styles.textLeft : styles.textRight)]}>
--- a/GiftedMessenger.js
+++ b/GiftedMessenger.js
@@ -241,7 +241,7 @@ var GiftedMessenger = React.createClass({
     this._rowIds = [];
     this.appendMessages(nextProps.messages);
 
-    let textInputHeight = this.props.style.textInputContainer.height || 44;
+    var textInputHeight = this.props.style.textInputContainer.height || 44;
 
     if (nextProps.maxHeight !== this.props.maxHeight) {
       this.listViewMaxHeight = nextProps.maxHeight;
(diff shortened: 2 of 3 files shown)


FaridSafi:
Thank you @bpeters, I will merge today

koppelaar:
@bpeters @FaridSafi Did you by any chance also build that the component deals with changes in maxHeight? In my project I'm updating maxHeight under various circumstances (e.g. when you rotate the screen), but since the variable is only processed during the initial render, it has no effect. This hides some of the content in the ListView for example after rotating the screen.

koppelaar:
I also noticed that the textInputHeight is hard coded, even though it can be overruled using the styles attribute. 

Currently this happens:

`var textInputHeight = 0;`
`if (this.props.hideTextInput === false) {`
`  textInputHeight = 44;`
`}`

Which would cause an issue when calling GiftedMessenger like this, for example:

`<GiftedMessenger`
`  style={{textInputContainer: {height: 60}}}`
`>`

The `textInputHeight` should probably be set at `this.props.style.textInputContainer.height` if it exists?

bpeters:
@koppelaar Good points, let me address them tonight and update the PR

FaridSafi:
I'll wait for next PR then

bpeters:
@koppelaar @FaridSafi Added those changes you suggested. Now the style height prop can be used as well as the max height prop to adjust the list height.

koppelaar:
@bpeters @FaridSafi Is this ready to be merged now?

PR #76, opened 2016-03-11
Add "typingText" prop and "typing" state to inform of typing

kesha-antonov:


wenkesj:
A couple things I see that could make this a bit more well-rounded IMO:
+ Do you think you could squash your commit for this merge?
+ Also, for the scroll fixes, can you please submit another PR?
+ Alternatively, to the above, could please add a more detailed explanation of the PR?

GiftedMessenger.js
```diff
@@ -240,6 +244,25 @@ var GiftedMessenger = React.createClass({
     this.appendMessages(nextProps.messages);
   },
 
+  componentWillUpdate(nextProps, nextState) {
+    if( this.state.typing !== nextState.typing ) {
+      switch( nextState.typing ) {
```
austinkelleher: I agree with @wenkesj. Perhaps something like this:

```javascript
return Animated.timing(this.state.typingOpacity, {
  toValue: nextState.typing ? 1 : 0,
  duration: 250,
  }).start();
};
```

FaridSafi:
A typingMessage prop has been implemented in 0.1.0

kesha-antonov:
Cool! Thanks.

PR #81, opened 2016-03-13
Add prop that causes names to show inside a bubble, rather than above it

koppelaar:
Many apps want to show names inside the text bubble, rather than above it. I've added a prop `displayNamesInsideBubble`, to enable this (by setting it to true). It only does something if `displayNames` is true as well. 

Ideally both of these props would be combined in a single prop `displayNames`, which could have values such as 'no', 'inside-bubble' and 'above-bubble', but since that might break previous implementations I have chosen to make it a separate prop.


koppelaar:
@FaridSafi would you be interested in merging this?

FaridSafi:
Looks good can you update also the readme please?

koppelaar:
Done!

PR #82, opened 2016-03-14
Fix keyboardDismissMode and add keyboardShouldPersistsTaps

### LINKED ISSUE #80: Tapping a single time on the Listview should dismiss the keyboard
The most common way for messenger apps to dismiss the keyboard is not through scroll interactions but by simply tapping the Listview (that contains the messages) once. GiftedMessenger does not do this, which makes it quite hard for users to dismiss the keyboard, I've noticed.

koppelaar:
The keyboardDismissMode was provided and documented, but wasn't actually used. This has now been restored/added. 

Also, I've exposed the keyboardShouldPersistsTaps prop of the ListView, so you can now choose whether or not the keyboard will be dismissed when you tap the ListView. This should resolve #80. I realise there is an underlying issue with this (keyboardShouldPersistTaps={false} + textInput focused = 2 taps are needed to trigger the ParsedText links), but IMHO it should be up to the person using GiftedMessenger to decide what weighs heavier. In my case, I really want to close the keyboard on tapping the ListView and am willing to accept the double press for opening links. 

@FaridSafi I hope you don't mind if I keep submitting some additions to your component? :)

### DIFF
--- a/GiftedMessenger.js
+++ b/GiftedMessenger.js
@@ -51,6 +51,8 @@ var GiftedMessenger = React.createClass({
       onImagePress: null,
       onMessageLongPress: null,
       hideTextInput: false,
+      keyboardDismissMode: 'interactive',
+      keyboardShouldPersistTaps: true,
       submitOnReturn: false,
       forceRenderImage: false,
       onChangeText: (text) => {},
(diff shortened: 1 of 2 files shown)


FaridSafi:
Thank you @koppelaar, PR are very welcome

PR #86, opened 2016-03-17
Message: add 'undefined' check to rowData.image

zxcpoiu:
sometimes we don't want to store/send redundant `rowData.image = null` if we don't need it.
add this check to prevent `Required prop 'source' was not specified in 'Image'`


FaridSafi:
thank you @zxcpoiu

PR #89, opened 2016-03-26
small things

Rusya13:
Check text for null or underfined before get length from it.
Use forEach instead of for … to work with Realm’s <Results>


FaridSafi:
Hello @Rusya13, thank you. What if message.id is not defined?

hsvlz:
Hello@Rusya13 Any chance you could share your additional development with Realm?

PR #91, opened 2016-04-04
Take maxHeight changes into account

Mokto:
This use case is need when rotating the screen.

Thanks.

### DIFF
--- a/GiftedMessenger.js
+++ b/GiftedMessenger.js
@@ -273,6 +273,10 @@ var GiftedMessenger = React.createClass({
       this.setState({
         height: new Animated.Value(this.listViewMaxHeight),
       });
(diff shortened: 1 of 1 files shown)


FaridSafi:
Hello @Mokto 
is your PR the same fix as https://github.com/FaridSafi/react-native-gifted-messenger/pull/92 ?

Mokto:
Oh, yes !

But I think there is a missing  - textInputHeight; (see my 2nd commit)

jevakallio:
I just hit the same issue and came to submit pull request with same changes as https://github.com/FaridSafi/react-native-gifted-messenger/pull/91/commits/8e73b5c884a846448c319a1e7e1dcab7dc8d8828 .

@Mokto @FaridSafi could we get the branch rebased and released? Would be very helpful!

PR #93, opened 2016-04-05
Adds Triangle next to bubble

swapkats:
screen shot 2016-04-05 at 17 05 59

### DIFF
--- a/Bubble.js
+++ b/Bubble.js
@@ -29,7 +29,17 @@ let styles = StyleSheet.create({
   bubbleError: {
     backgroundColor: '#e01717'
   },
-
+  triangleCorner: {
+    position: 'absolute',
+    width: 0,
+    height: 0,
+    backgroundColor: 'transparent',
+    borderStyle: 'solid',
+    borderRightWidth: 10,
+    borderTopWidth: 10,
+    borderRightColor: 'transparent',
+    borderTopColor: '#e6e6eb'
+  }
 });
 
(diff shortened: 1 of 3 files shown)


FaridSafi:
Hello, thank you @swapkats it was a requested feature!
Can you add a prop to enable the triangles please?

braco:
Why not achieve this with something more flexible, like children?:

https://github.com/FaridSafi/react-native-gifted-messenger/issues/129

or the ability to pass in custom View elements to replace the core Classes. That way, you could just swap in elements:

```
import Bubble from 'react-native-gifted-messenger/bubble-speech'
<GiftedMessenger bubbleView={Bubble} />
```

swapkats:
@FaridSafi Something like 0f94db6fd4cc1813731b94c83d09d080b4e10522 ?

@braco It is neater on my end to use gifted messenger if i don't do that, one less import to worry about

ncnlinh:
@swapkats I think this is awesome! BTW, you should also add README for this.

swapkats:
@ncnlinh I did https://github.com/FaridSafi/react-native-gifted-messenger/pull/93/files?short_path=04c6e90#diff-04c6e90faac2675aa89e2176d2eec7d8

mywaystar:
Any news?

mywaystar:
Note that you should add flexStyle to the root view to wrap text

PR #104, opened 2016-04-21
Replace lodash with deep-equal.

austinkelleher:
lodash is a large unnecessary dependency when only using a few small functions. Also, the current version on npm is broken because the newest version with [this fix](https://github.com/FaridSafi/react-native-gifted-messenger/commit/bda7ac8fe5bf41af5539cb200ab14e1aa7ae4918) has not been published.


wenkesj:
:+1:

wenkesj:
@FaridSafi lgtme

FaridSafi:
Thank you @austinkelleher

PR #109, opened 2016-04-21
Add dateLocale prop to set the correct moment.js locale

gnl:
...for the date/time labels

It's an ugly solution, but dynamically assembling the string passed to require doesn't work, sadly.


FaridSafi:
Yes it's verbose but no alternative.
Thank you @gnl

PR #116, opened 2016-05-03
Multiline TextInput + Initials avatar support

faceyspacey:
To enable multiline textinputs, set `multiline` to true as a property to `GiftedMessenger` to enable. set `maxInputHeight` to specify the maximum height the input can grow too. If you need the change event for when it changes height, use `onTextInputHeightChanged`. You can pass additional props to the input now with `textInputProps`. 

IMPLEMENTATION NOTES:
The autogrowing part was easy. What was very time-consuming was getting the positioning compensation for the ListView so the correct part was always placed properly (similar to the iPhone text message app). I had to animate the textinput separately (well, its container). So similar to the ListView container, I animate its height. They are both done in parallel, but could likely be optimized to interpolate. The reason it had to be done separately--and believe me, I tried to use the initial mechanism to move the input up--is because when the placement of the textinput is dependent on the height of the ListView, then you can't move the ListView (in relation to the TextInput) without also moving the TextInput. I spent a long time trying to do that via various ways and things just got tangled, so I did what I probably should have originally did which was isolate the TextInput so it moves in an unrelated fashion. Absolute positioning and a `bottom` property are used to have the textinput stick to the bottom and grow up from the bottom. Aside from the listview placement compensation, the bottom-based absolute positioning is its secret sauce.

### DIFF
--- a/GiftedMessenger.js
+++ b/GiftedMessenger.js
@@ -16,6 +16,7 @@ import moment from 'moment';
 import {setLocale} from './Locale';
 import deepEqual from 'deep-equal';
 import Button from 'react-native-button';
+import {AutoGrowingTextInput} from 'react-native-autogrow-textinput';
 
 class GiftedMessenger extends Component {
 
@@ -42,12 +43,16 @@ class GiftedMessenger extends Component {
     this._scrollToPreviousPosition = false;
     this._visibleRows = { s1: { } };
 
+    this.totalInputHeightDelta = 0;
     let textInputHeight = 44;
+
     if (!this.props.hideTextInput) {
       if (this.props.styles.hasOwnProperty('textInputContainer')) {
         textInputHeight = this.props.styles.textInputContainer.height || textInputHeight;
       }
     }
+
+    this.textInputHeight = textInputHeight;
 
     this.listViewMaxHeight = this.props.maxHeight - textInputHeight;
 
@@ -65,6 +70,7 @@ class GiftedMessenger extends Component {
       text: props.text,
       disabled: true,
       height: new Animated.Value(this.listViewMaxHeight),
+      inputHeight: new Animated.Value(this.textInputHeight),
       appearAnim: new Animated.Value(0),
     };
   }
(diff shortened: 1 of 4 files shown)


yanghuxiao:
good job!

basdvries:
@FaridSafi Any indication when this can be merged/will be reviewed?

GiftedMessenger.js
```diff
@@ -648,7 +648,7 @@ class GiftedMessenger extends Component {
   //of the page if the height taken up by the messages is less than the height available, but if the messages take up more than the available
   //height, the bottom of the messages should align with the top of the textinput. See `this.hasMoreContentThanSpaceAvailable()`.
   onTextInputHeightChanged(newHeight, oldHeight, heightDelta) {
-    if(heightDelta > -5 && heightDelta < 5) return; //after typing the first character in the input, a small height increase occurs, which isn't compensating by a height decrease when it's deleted, so we disregard it
+    if(heightDelta > -10 && heightDelta < 10) return; //after typing the first character in the input, a small height increase occurs, which isn't compensating by a height decrease when it's deleted, so we disregard it
```
wenkesj: I think this should be a constant rather than hard coded, this has been changed a couple times. It seems like it could be more useful to have better supporting reasoning.

README.md
```diff
@@ -81,7 +81,12 @@ See [GiftedMessengerExample/GiftedMessengerContainer.js](https://raw.githubuserc
 | senderName                    | String   | Name of the sender of the messages                                         | Both     | 'Sender'                         |
 | styles                        | Function | Styles of children components - See GiftedMessenger.js/componentWillMount  | Both     | {}                               |
 | submitOnReturn                | Boolean  | Send message when clicking on submit                                       | Both     | false                            |
-| typingMessage                 | String   | Display a text at the bottom of the list. Eg: 'User is typing a message'   | Both     | ''                               |
+| typingMessage                 | String   | Display a text at the bottom of the list. Eg: 'User is typing a message'   | Both     | ''    
+| textInputProps                | Object   | Props passed to Textinput                                                  | Both     | {}  
+| multiline                     | Boolean  | When true, textinput expands vertically on return                          | Both     | false
```
wenkesj: Should `multiline` be here? TextInput props include `multiline`, maybe to avoid confusion make a note here. Not too important.

package.json
```diff
@@ -33,7 +33,8 @@
     "moment": "^2.10.6",
     "react-native-button": "^1.3.1",
     "react-native-gifted-spinner": "0.0.3",
-    "react-native-parsed-text": "0.0.11"
+    "react-native-parsed-text": "0.0.11",
+    "react-native-autogrow-textinput": "github:faceyspacey/react-native-autogrow-textinput"
```
wenkesj: I'm not sure about this. There isn't any added security for this; can you register this on NPM? If you decide for some reason to change/delete the repo, it's gone!

wenkesj:
@faceyspacey, This could be cleaned up a little bit, but I would like to see this merged as well. Thanks!

basdvries:
@faceyspacey while you're at it, i'd also suggest to submit multiple pulls with the features split up. Let me know if you're going to do a refactoring. If not I'll take care of some of the features pulled for here.

faceyspacey:
Hey guys, I've reached the stage of my project where we're about to launch and I only have time for urgent stuff. I can't forecast when I'll have time for this. 

....as far as the AutoGrow repo required, its creator basically made the requested changes. So we should be able to just use his repo now. There may be one modification we want to make to our code though.

PR #125, opened 2016-05-16
renderCustomText render if not return false value  and fix two-byte characters length error

yanghuxiao:
in this issues:https://github.com/FaridSafi/react-native-gifted-messenger/issues/124,i want to render my custom text.then now i can use:
rederCustomText(rowData){
if(rowData.type=='image'){
....reder CustomText
} else {
   return false;
}
}


yuanhsh:
LGTM. @FaridSafi Can you merge this PR and bump the version?

FaridSafi:
Thank you guys

PR #131, opened 2016-05-27
Wait with animation till keyboard will fully rendered

alimek:
Hi, 

Just copy&pasted your example and when I tried to load the scene with that code, It gives me error, that resolve the issue 👍


alimek:
@FaridSafi what you think about this PR? Is it sill valid to merge it?

PR #133, opened 2016-05-31
Display 1 avatar per "cluster" of messages from the same user

nlively:
As a chat user I find it annoying to see one avatar per message if I have just sent 5 messages in sequence.

This modification to the code allows messages to be grouped better such that avatars are displayed less frequently.

### DIFF
--- a/GiftedMessenger.js
+++ b/GiftedMessenger.js
@@ -42,7 +42,7 @@ class GiftedMessenger extends Component {
     this._footerY = 0;
     this._scrollToBottomOnNextRender = false;
     this._scrollToPreviousPosition = false;
-    this._visibleRows = { s1: { } };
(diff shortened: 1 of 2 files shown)


GiftedMessenger.js
```diff
@@ -2,6 +2,7 @@ import React, {
   Component,
 } from 'react';
 import {
+  ActivityIndicatorIOS,
```
sibeliusseraphini: it is better to use a cross platform component like - https://github.com/FaridSafi/react-native-gifted-spinner

`ActivityIndicatorIOS` will be replace by `ActivityIndicator` on RN0.28

fukemy:
hi, i still got multi image for same user, how did i got this feature? Thanks

PR #159, opened 2016-07-13
[FEATURE] Add background image support

nicoecheza:
Allow gifted messenger to have a background image :dancer:

### DIFF
--- a/GiftedMessenger.js
+++ b/GiftedMessenger.js
@@ -7,6 +7,7 @@ import {
   ListView,
   TextInput,
   Dimensions,
+  Image,
   Animated,
   Platform,
   PixelRatio,
@@ -79,6 +80,12 @@ class GiftedMessenger extends Component {
       },
       listView: {
         flex: 1,
+      },
+      backgroundImage: {
+        flex: 1,
+        width: null,
+        alignSelf: 'stretch',
+        height: null,
       },
       textInputContainer: {
         height: 44,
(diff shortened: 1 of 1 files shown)


a-sierra:
nice work!

manueloteroeiras:
is a great feature, tks @nicoecheza

nicoecheza:
Also allow for custom send button rendering

GiftedMessenger.js
```diff
@@ -647,6 +669,7 @@ GiftedMessenger.defaultProps = {
 
 GiftedMessenger.propTypes = {
   autoFocus: React.PropTypes.bool,
+  backgroundImage: React.PropTypes.number,
```
laoshanlung: Should/can this be either a number (ID of the StyleSheet) or an object referring to the style itself?

nicoecheza: This should be the imported image, so you should give th GiftedMessenger a prop like:
`backgroundImage={ require("image.png")`

laoshanlung: Opps, I misread it, thought it was the style `backgroundImage`. Anyway, should it be a string in case someone wants to pass an URL?

nicoecheza: I thought about that, but I don't think that passing an URL is a good idea since it will be a pretty large background image and the waiting times for it would be very slow considering 3G or 4G connections.
Anyway, that's not up to me, so I will be updating this to also accept URLs :dancer:

guns2410:
Not sure why this is not merged yet?

greek-stasia:
Can we merge this please

PR #160, opened 2016-07-15
Added onfocus handler

jimmyhmiller:
I need this in order to make the messenger go full screen when you click on the textbox. If I don't do that, the scroll does not work properly in the card ui.


jimmyhmiller:
Sorry, that was supposed to be to my personal one.

PR #173, opened 2016-08-01
Move md5 hash to message instead of all messages

kfiroo:


kfiroo:
@FaridSafi what do you think?

src/MessageContainer.js
```diff
@@ -1,29 +1,50 @@
-import React, { Component, PropTypes } from 'react';
+import React, {Component, PropTypes} from 'react';
+
+import ReactNative, {
+  View,
+  ListView,
+  UIManager
```
FaridSafi: is UIManager used ?

src/MessageContainer.js
```diff
@@ -47,46 +68,64 @@ export default class MessageContainer extends Component {
...
+      </View>
     );
   }
 }
 
 MessageContainer.defaultProps = {
   messages: [],
   user: {},
-  footer: null,
+  renderFooter: null,
```
FaridSafi: I guess it should also be a default props of GiftedChat.js

src/MessageContainer.js
kfiroo: Nope, I tried to make the content stick to the top when its smaller then the container, it didn't work that well so I removed it.
I will remove the left over reference

FaridSafi:
I will do the following tests with a device :

While having 400 messages displayed
- Typing a character should feels reactive
- Typing a character in the text input should not trigger ListView rowHasChanged call

src/MessageContainer.js
kfiroo: fixed

FaridSafi:
Typing a character in the text input should not trigger ListView rowHasChanged call
=> rowHasChanged is called for each character typed

kfiroo:
Ok, will check

kfiroo:
@FaridSafi does it return `true` for each row?

FaridSafi:
it returns `false`

kfiroo:
@FaridSafi Then we are good :)
The list view will attempt to render the rows only if `rowHasChanged` returns `true`

FaridSafi:
yes but it slows the typing

kfiroo:
I will try to fix it

FaridSafi:
What do you think about keeping the hash at the messages list level and also at the message level ?

FaridSafi:
`initialListSize={this.props.messages.length}` is needed to display more than 22 rows
I guess `pageSize` will be needed also, it should be the amount of displayable rows in the screen (without scrolling)

I have to say that I am not convinced yet by the use of ListView

FaridSafi:
rowHasChanged calls for each typed character can create this warning

simulator screen shot 01 aug 2016 16 47 53

kfiroo:
@FaridSafi ListView is supposed to handle large sets of data much better then static rendering, yet for now I agree this is not stable enough to merge.
I will play around with it and let you know if I got anything interesting :)

kfiroo:
@FaridSafi closing this and submitting a new one without the unneeded renders

Issue #175, opened 2016-08-02
RenderMessageImage does nothing


explorigin:
The fix should be as easy as adding a return at: https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/Bubble.js#L36

FaridSafi:
Thank you @explorigin

PR #176, opened 2016-08-02
Fix renderMessageImage/renderMessageText not overriding the default.

explorigin:
Addresses issue #175


FaridSafi:
Thank you! :)

PR #178, opened 2016-08-02
Move hash to single message from all messages

kfiroo:
Moving hash to each message should allow MessageContainer to rerender when messages are not changed.
Use ListView with InvertibleListView to easily support a lot of messages and have renderFooter, renderHeader etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "@exponent/react-native-action-sheet": "git+https://github.com/FaridSafi/react-native-action-sheet.git",
     "md5": "^2.1.0",
     "moment": "^2.13.0",
+    "react-addons-shallow-compare": "^15.2.1",
     "react-native-communications": "^2.0.0",
     "react-native-dismiss-keyboard": "^1.0.0",
     "react-native-invertible-scroll-view": "^1.0.0",
--- a/src/GiftedChat.js
+++ b/src/GiftedChat.js
@@ -275,7 +275,7 @@ class GiftedChat extends Component {
 
           ref={component => this._messageContainerRef = component}
         />
-        {this.renderFooter()}
+        {this.renderChatFooter()}
       </AnimatedView>
     );
   }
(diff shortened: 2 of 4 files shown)


FaridSafi:
Thanks for you work @kfiroo 
It seems that messagesHash has better performance than an hash for each messages

Using the following data https://gist.github.com/FaridSafi/803bf49b6fa1f82709dc9c9a5462b861
I appended a new message and it took 3-4 seconds versus 1 second using messagesHash

Note :
ListView should have also
```
          initialListSize={this.props.messages.length}
          pageSize={15} // maybe more ?
```

kfiroo:
Actually `initialListSize` should not be too big, it only indicates how many rows the `ListView` should render in the first render cycle, if that number is too big it will cause the UI thread to hang and the app will feel stuck.
We should render a number that will fit in the screen so the user won't feel the batched loading, `ListView` will take care of loading the others in cycles.

I really think `ListView` is the better solution, rendering a lot of data inline is really discouraged by RN team.
`ListView` also supports native view recycling and not removing views that are not in the viewport to enhance performance.

I will try to improve the `append` performance and I'll ping you again :)

kfiroo:
@FaridSafi Hey :)
So, performance seems to be very good now,
(**note all measurements are in DEV mode so it should be significantly better in production**)
* List only renders `initialListSize` messages so first render takes ~200ms
* Lists with under 100 items `append` takes ~50ms
* Lists with 200 items `append` takes ~100ms
* Lists with 400 items `append` takes ~250ms

Let me know what you think

kfiroo:
There is a bug with previous/next message reference - fixing it now

FaridSafi:
I am testing today :)

FaridSafi:
Thank you @kfiroo
I confirm that appending a message is faster!

I will try to replace `react-addons-shallow-compare` by `shallowequal` to avoid a warning during the installation

FaridSafi:
I released 0.0.4 with your updates. Good work!
Thanks

kfiroo:
@FaridSafi Great! Thanks

PR #180, opened 2016-08-05
add more XML context to android installation

calebmer:
It’s hard to understand where exactly the attribute goes without more context.


FaridSafi:
thank you!

Issue #169, opened 2016-07-31
Keyboard height on iOS


TimBroddin:
First: congrats on the release of the new component! Looks like a huge improvement over gifted-messenger!

I'm having trouble with the keyboard leaving a gap (same height as my navbar it seems):

<img width="374" alt="schermafbeelding 2016-07-31 om 15 24 17" src="https://cloud.githubusercontent.com/assets/528287/17276767/1b1764a6-5733-11e6-9536-b4a9fb7b49a5.png">

<img width="375" alt="schermafbeelding 2016-07-31 om 15 24 25" src="https://cloud.githubusercontent.com/assets/528287/17276770/1dad6684-5733-11e6-968a-5c8dd69a6c1c.png">

Any idea on how to fix this? (The navigation I'm using is react-native-navigation)

Thanks

FaridSafi:
Hello @TimBroddin,
Does your GiftedChat container has a paddingBottom ? If yes try to switch to a marginBottom

FaridSafi:
What if you wrap GiftedChat inside a `<View style={{flex: 1}} />` ?
I tried to set up the react-native-navigation but the example is in RN 0.25.1

I will need a code sample to help you more

ssmlee04:
same here. Mine is RN 0.28.0. Keyboard size is about the same as the one above.

FaridSafi:
@ssmlee04 do you use react-native-navigation?

Can you share part of your code guys?

ssmlee04:
I used <TabBarIOS> and a <AdMobBanner>. And it seems the extra height equals to the sum of the two. I'll see if I can come up with something minimum.

FaridSafi:
@ssmlee04 Did you try to wrap GiftedChat inside a <View style={{flex: 1}} /> ?

TimBroddin:
@FaridSafi I'll try to make an example repo tonight

FaridSafi:
@TimBroddin I can reproduce the issue, I will fix it

TimBroddin:
Sorry it took so long @FaridSafi, I made an example repository: https://github.com/TimBroddin/gifted-chat-keyboard-bug

FaridSafi:
I added `bottomOffset` prop in 0.0.5
You need to set the distance from the bottom of the screen

```js
<GiftedChat bottomOffset={56} />
```

TimBroddin:
Works, thanks a lot @FaridSafi!

[...]

PR #187, opened 2016-08-12
Changed direct use of the bottomOffset prop into getter / setter, all…

ckuijjer:
Changed direct use of the bottomOffset prop into getter / setter, allowing the bottomOffset only to be used when the keyboard is shown. Fixed initial render issue.

### DIFF
--- a/src/GiftedChat.js
+++ b/src/GiftedChat.js
@@ -42,6 +42,7 @@ class GiftedChat extends Component {
     // default values
     this._isMounted = false;
     this._keyboardHeight = 0;
+    this._bottomOffset = 0;
     this._maxHeight = null;
     this._touchStarted = false;
     this._isTypingDisabled = false;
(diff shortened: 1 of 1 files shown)


ckuijjer:
See #169

FaridSafi:
Thank you, I'll test tomorrow

FaridSafi:
I merged your PR manually https://github.com/FaridSafi/react-native-gifted-chat/commit/2133625d9befd4f85be958272470fc522fa9c236

FaridSafi:
I tested on Android and it's not enough to handle bottom views properly.
You will also need to hide the views below the chat when the keyboard appears.

Feel free to submit a PR for Android

ckuijjer:
I might take a look at the Android issue, seems like a good reason to finally play with the Android development tools.

PR #193, opened 2016-08-14
Allow pass 'multiline:false' to 'Composer' component

fengliu222:
I try to pass render custom composer param.
When i pass `multiline:false` to the composer , i got a red box.

'e.nativeEvent.contentSize' is not exist when `multiline` param set to false.
So i fix this problem. ( oh my pool english ........)


fengliu222:
usage:

```jsx
  _renderComposer(props) {
    return (
      <Composer
                {...props}
                textInputProps={{
                  returnKeyType: 'send',
                  multiline: false,
                  onSubmitEditing : (event) => {
                    props.onSend({ text : event.nativeEvent.text.trim() }, true);
                  }
                }}
      />
    )
  }
```

FaridSafi:
Thank you, I'll test tomorrow

kfiroo:
This PR now has 2 separate fixes, for `multiline` and for `renderAvatar`.
Maybe we should split it into two PRs?

FaridSafi:
Looks good, thank you!

Issue #185, opened 2016-08-11
[Request] Add prop for tintColor for ActionSheet in Actions


devbriguy:

kfiroo:
@devbriguy PRs are welcome :)

devbriguy:
PR was merged

PR #197, opened 2016-08-15
[Request] Add prop for tintColor for ActionSheet in Actions #185

devbriguy:
Add prop optionTintColor for ActionSheet options text


FaridSafi:
Thank you for this PR but the default color 'blue' is not the native one

devbriguy:
yup you're right, it's #007AFF. I'll make the change.

devbriguy:
Just checking in, is this solution not ideal?

PR #202, opened 2016-08-17
add onPress event

kesha-antonov:


FaridSafi:
Thank you!

FaridSafi:
I changed a bit the API https://github.com/FaridSafi/react-native-gifted-chat/commit/ba543061e68d2cd7c44784efe16fd7bee5ff5d8f

To use the onPress event you will have to do

```jsx
import {GiftedChat, Bubble} from 'react-native-gifted-chat';

<GiftedChat renderBubble={(props) => {
    return (
      <Bubble
        {...props}
        touchableProps={{
          onLongPress: null,
          onPress: () => {
            alert('onPress');
          },
        }}
      />
    );
}}
```

but I recommend to use `onLongPress` instead of `onPress`

kesha-antonov:
Thanks!

I need to use onPress and onLongPress together for different actions

manjeet-dobble:
@FaridSafi I tried as suggested by you , but onPress don't seems working, I also checked source for Bubble.js and there is no onPress function as shown in this [link](https://github.com/FaridSafi/react-native-gifted-chat/commit/ba543061e68d2cd7c44784efe16fd7bee5ff5d8f)

kfiroo:
@manjeet-dobble the `onPress` is set [here](https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/Bubble.js#L105)

PR #203, opened 2016-08-17
add day format as calendar

kesha-antonov:
I saved format of current version of Day and added ability to pass dayFormat to use in calendar() of moment.

Hence we can have 'TODAY', 'YESTERDAY' etc. Instead of '16 Aug. 2016' when it was yesterday

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,7 @@ See [example/App.js](example/App.js)
 - **`user`** _(Object)_ - user sending the messages `{_id, name, avatar}`
 - **`onSend`** _(Function)_ - function to call when sending a message
 - **`locale`** _(String)_ - localize the dates
+- **`dayFormat`** _(Object)_ - format for day. `{sameDay, nextDay, lastDay, nextWeek, lastWeek, sameElse}`. Can be used with locale. `{en: {sameDay,...}, ru: {sameDay,...}}`. Pass `null` to have default output of calendar function of momentjs.
 - **`isAnimated`** _(Bool)_ - animates the view when the keyboard appears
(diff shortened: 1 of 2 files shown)


FaridSafi:
Hi!
I tried this before, the issue is that `moment` supports 'TODAY', 'YESTERDAY' only in english.

kesha-antonov:
I disagree with you. Now it supports in many languages. Well, russian and english at least
https://github.com/moment/moment/blob/develop/src/locale/ru.js#L73

kesha-antonov:
I bet and other languages too.
Clicked random and saw it: https://github.com/moment/moment/blob/develop/src/locale/pl.js#L55

FaridSafi:
for example in french it's "Aujourd'hui à 15:00"
I would like to display only the day since the time is already displayed inside the bubble

kesha-antonov:
Ok. Got it.

So now you can copy from https://github.com/moment/moment/blob/develop/src/locale/fr.js#L23
```
        sameDay: '[Aujourd\'hui à] LT',
        nextDay: '[Demain à] LT',
        nextWeek: 'dddd [à] LT',
        lastDay: '[Hier à] LT',
        lastWeek: 'dddd [dernier à] LT',
        sameElse: 'L'
```

and pass it to Chat like this:
```
dayFormat: {
        sameDay: '[Aujourd\'hui]',
        nextDay: '[Demain]',
        nextWeek: 'dddd',
        lastDay: '[Hier]',
        lastWeek: 'dddd [dernier]',
        sameElse: 'L'
}
```

I don't know French so correct me if I have typos here :)

PR #204, opened 2016-08-17
add load earlier label

kesha-antonov:


FaridSafi:
Hello @kesha-antonov,
You can already do :
```jsx
import { GiftedChat, LoadEarlier } from 'react-native-gifted-chat';

<GiftedChat
  renderLoadEarlier={(props) => {
    return (
      <LoadEarlier {...props} label='Custom Label' />
    );
  }}
/>
```

kesha-antonov:
Ok thanks! Now I learned "the new way" :)

PR #206, opened 2016-08-17
add on type

kesha-antonov:


kfiroo:
@kesha-antonov @FaridSafi I really don't like the fact that `onType` handles text changes as well as layout changes, it puts a lot of unrelated logic in one place.
As a user I would probably like to subscribe to text change event to update some state,
OR subscribe to layout changes to update some other layout.

When this logic is contained within the component it is "ok" but exposing it feels wrong :/

kesha-antonov:
@kfiroo But code is used in GiftedChat "onType" event
What's wrong? https://github.com/kesha-antonov/react-native-gifted-chat/blob/80e2015a03a8949ab5bea99cbbe4fbee7d6cc4f5/src/GiftedChat.js#L348

If you can please improve my code :) I need this event.

kfiroo:
@kesha-antonov Can you share your use case please? :)

FaridSafi:
You can also do something like this
```jsx
import {GiftedChat, Composer} from 'react-native-gifted-chat';

<GiftedChat
  renderComposer={(props) => {
    return (
       <Composer {...props} onChange={(e) => {
         // onType logic here
         props.onChange(e);
       }} />
    );
  }}
/>
```

kesha-antonov:
@FaridSafi Hmm seems you have all i need :) I'll try that and then close pull

kesha-antonov:
Yep. It worked. Thanks, Farid!

PR #207, opened 2016-08-17
fix footer re-render

kesha-antonov:
Hello!

I've got a
```
renderFooter: ->
    if @state.typingText?
      <View style={styles.footerContainer}>
      ...
    else
      null
```
function. But it's not triggered when I update `typingText` in `state`.
Only when messages replaced/updated.

IMO it's wrong behaviour.
But maybe you can tell some advices/thoughts if this PR is right?

### DIFF
--- a/src/MessageContainer.js
+++ b/src/MessageContainer.js
@@ -62,7 +62,7 @@ export default class MessageContainer extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    if (this.props.messages === nextProps.messages) {
+    if (this.props.messages === nextProps.messages && this.props === nextProps) {
       return;
     }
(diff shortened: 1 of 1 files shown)


kfiroo:
@kesha-antonov The line you changed in `componentWillReceiveProps` only determines if we need to update the `dataSource`, and we only want to do so when the messages change as this might be expensive (correct me if I'm wrong)

As a temp solution you can pass the `typingText` state as a prop to `GiftedChat` which will cause it to re-render when it is changed and thus calling your `renderFooter` function.

kesha-antonov:
Thanks for suggestion! I'll try that! @kfiroo

kesha-antonov:
Thanks @kfiroo ! It worked.

PR #213, opened 2016-08-22
Fix unnecessary warning when message.id === 0

cooperka:
`0` evaluates to `false`, so this warning will appear for anyone using a 0-based array index for their messages.

Alternative syntax could be `if (!(message._id || message._id === 0))`, if you have any particular preference for readability.


FaridSafi:
Thank you @cooperka!

PR #217, opened 2016-08-25
Add accessibilityTraits to support screen readers

cooperka:
Using `accessibilityTraits="button"`, for example, allows screen readers like Voiceover or TalkBack to behave more appropriately. See the [React Native](https://facebook.github.io/react-native/docs/accessibility.html) or [Apple](https://developer.apple.com/library/ios/documentation/UserExperience/Conceptual/iPhoneAccessibility/Making_Application_Accessible/Making_Application_Accessible.html#//apple_ref/doc/uid/TP40008785-CH102-SW7) guides for more info.

### DIFF
--- a/src/Bubble.js
+++ b/src/Bubble.js
@@ -101,6 +101,7 @@ export default class Bubble extends React.Component {
         <View style={[styles[this.props.position].wrapper, this.props.wrapperStyle[this.props.position], this.handleBubbleToNext(), this.handleBubbleToPrevious()]}>
           <TouchableWithoutFeedback
             onLongPress={this.onLongPress}
+            accessibilityTraits="text"
             {...this.props.touchableProps}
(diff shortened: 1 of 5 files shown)


FaridSafi:
Thank you @cooperka! I like this new feature

I am just curious about 2 things :
- Why Bubble instead of MessageText and MessageImage?
- What about Send button?

cooperka:
Good catch! I tested everything except for actually typing a message 😎. I've updated Send button too.

I added to Bubble because currently that's the component that actually receives the touch event when you tap it (see screenshot).

simulator screen shot aug 25 2016 17 53 18

FaridSafi:
thank you!

cooperka:
Thank you! I appreciate your quick responses 🍻

PR #220, opened 2016-09-01
Updating font to avenir

jakehasler:


kfiroo:
@jakehasler I feel like setting the font should be a customisable to the user and not hard coded
What do you think?

paramaggarwal:
Will this even work on Android? This feels wrong to me.

FaridSafi:
Hi, you set a custom font using `renderMessage`

jakehasler:
@kfiroo @FaridSafi I apologize for the accidental PR here. I meant to perform the change to a forked version. Thanks for the `renderMessageText` input - I'll be sure to implement that. Thanks for a great module!

PR #221, opened 2016-09-01
add prop hideInputToolbar

kesha-antonov:


kfiroo:
@kesha-antonov Hey,
Wouldn't using `renderInputToolbar` to return `null` or an empty `View` have the same result?

kesha-antonov:
@kfiroo Hey! Thanks, I'll try it!

kesha-antonov:
@kfiroo No. When I return ```null/<View/>``` in ```renderInputToolbar``` it is blank but height is not 0.
screen shot 2017-01-20 at 13 33 54

kfiroo:
We need to think this through. The proper way to resolve this, in my opinion, is to have the `InputToolbar` dynamically resize according to its content.
Not sure why this isn't the case right now, maybe there are some technical difficulties?

PR #226, opened 2016-09-07
Initial support tick messages

jeromegrosse:
I thought it would be a good idea to support ticks appearing in the bubble ala WhatsApp or Telegram depending if the message has been delivered and read.

Here is a screenshot of how it appears as of now with both the read and delivered flag set to `true`:

screenshot_20160907-164845

What do you think?

### DIFF
--- a/src/Bubble.js
+++ b/src/Bubble.js
@@ -64,8 +64,8 @@ export default class Bubble extends React.Component {
     if (currentMessage.sent || currentMessage.received) {
       return (
         <View style={styles.tickView}>
-          {currentMessage.sent && <Text style={styles.tick}>✓</Text>}
-          {currentMessage.received && <Text style={styles.tick}>✓</Text>}
+          {currentMessage.sent && <Text style={[styles.tick, this.props.tickStyle]}>✓</Text>}
+          {currentMessage.received && <Text style={[styles.tick, this.props.tickStyle]}>✓</Text>}
         </View>
       )
     }
@@ -216,6 +216,7 @@ Bubble.defaultProps = {
   previousMessage: {},
   containerStyle: {},
   wrapperStyle: {},
+  tickStyle: {},
   containerToNextStyle: {},
   containerToPreviousStyle: {},
 };
@@ -241,6 +242,7 @@ Bubble.propTypes = {
     left: View.propTypes.style,
     right: View.propTypes.style,
   }),
+  tickStyle: Text.propTypes.style,
   containerToNextStyle: React.PropTypes.shape({
     left: View.propTypes.style,
     right: View.propTypes.style,
--- a/src/Bubble.js
+++ b/src/Bubble.js
@@ -54,14 +54,15 @@ export default class Bubble extends React.Component {
   }
 
   renderTicks() {
-    _renderTick = function(displayTick) {
-      if (displayTick) return <Text style={styles.tick}>✓</Text>;
-    }
     if (this.props.currentMessage.sent || this.props.currentMessage.received) {
+      const {currentMessage} = this.props;
+      if (this.props.renderTicks) {
+        return this.props.renderTicks(currentMessage);
+      }
       return (
         <View style={styles.tickView}>
-          {_renderTick(this.props.currentMessage.sent)}
-          {_renderTick(this.props.currentMessage.received)}
+          {currentMessage.sent && <Text style={styles.tick}>✓</Text>}
+          {currentMessage.received && <Text style={styles.tick}>✓</Text>}
         </View>
       )
     }
--- a/src/Bubble.js
+++ b/src/Bubble.js
@@ -54,11 +54,15 @@ export default class Bubble extends React.Component {
   }
 
   renderTicks() {
-    if (this.props.currentMessage.sent || this.props.currentMessage.received) {
-      const {currentMessage} = this.props;
-      if (this.props.renderTicks) {
+    if (this.props.renderTicks) {
         return this.props.renderTicks(currentMessage);
-      }
+    }
+
+    const {currentMessage} = this.props;
+    if (currentMessage.user._id !== this.props.user._id) {
+        return;
+    }
+    if (currentMessage.sent || currentMessage.received) {
       return (
         <View style={styles.tickView}>
           {currentMessage.sent && <Text style={styles.tick}>✓</Text>}
@@ -125,8 +129,8 @@ export default class Bubble extends React.Component {
               {this.renderMessageImage()}
               {this.renderMessageText()}
               <View style={styles.bottom}>
-                {this.renderTicks()}
                 {this.renderTime()}
+                {this.renderTicks()}
               </View>
             </View>
           </TouchableWithoutFeedback>
@@ -177,15 +181,16 @@ const styles = {
   }),
   bottom: {
     flexDirection: 'row',
+    justifyContent: 'flex-end',
   },
   tick: {
     fontSize: 10,
     backgroundColor: 'transparent',
-    color: 'white'
+    color: 'white',
   },
   tickView: {
     flexDirection: 'row',
-    marginLeft: 10,
+    marginRight: 10,
   }
 };
 
--- a/src/Bubble.js
+++ b/src/Bubble.js
@@ -54,11 +54,10 @@ export default class Bubble extends React.Component {
   }
 
   renderTicks() {
+    const {currentMessage} = this.props;
     if (this.props.renderTicks) {
         return this.props.renderTicks(currentMessage);
     }
-
-    const {currentMessage} = this.props;
     if (currentMessage.user._id !== this.props.user._id) {
         return;
     }
--- a/src/Bubble.js
+++ b/src/Bubble.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import {
+  Text,
   Clipboard,
   StyleSheet,
   TouchableWithoutFeedback,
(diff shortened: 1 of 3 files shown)


kfiroo:
@jeromegrosse Why are `messageRead` and `messageDelivered` props of `GiftedChat`?
Isn't it a per message state and not global?

jeromegrosse:
@kfiroo Yup, you are definitely right. I messed up there, I should use the messages array for that. I'll update the pull request with those changes as soon as possible.

jeromegrosse:
@kfiroo I've just pushed a few changes:
* The `messageRead` and `messageDelivered` props are no longer passed to GiftedChat as such but are instead set on a per message basis using the message state
* Also, a message is not *read* or *delivered* but rather *sent* and *delivered* so as to be closer to the concept of the ticks in other mobile instant messaging software

What do you think?

kfiroo:
@jeromegrosse Great! :)
I still have a few questions - I add them as inline comments

src/Bubble.js
```diff
@@ -52,6 +53,20 @@ export default class Bubble extends React.Component {
     return null;
   }
 
+  renderTicks() {
+    _renderTick = function(displayTick) {
```
kfiroo: Creating nested methods is very wasteful and not necasery in this case, so I think `_renderTick` should be a method of Bubble, or even as a stateless component.
Also looks like it is a global variable - did you forget the `const`?

kfiroo:
@jeromegrosse I'm missing a way to override the `renderTicks` method, If the user wants to change the style or behaviour.
But it might be good enough for the first commit, @FaridSafi what do you think?

FaridSafi:
I'll take a look this weekend

jeromegrosse:
@kfiroo
Thanks a lot for your feedbacks!

I performed the following modifications:
* Deletion of the nested method `_renderTicks`
* Added the possibility to override the way the ticks are rendered

However, I have a couple of questions:
* Should I rather put the ticks on the right of the time?
* Should I restrict the ticks to only the message sent by the user and never display the ticks for the received message?

src/Bubble.js
```diff
@@ -52,6 +53,21 @@ export default class Bubble extends React.Component {
     return null;
   }
 
+  renderTicks() {
+    if (this.props.currentMessage.sent || this.props.currentMessage.received) {
+      const {currentMessage} = this.props;
+      if (this.props.renderTicks) {
```
kfiroo: This could go above the first `if`
The user might want to have different props for his sent / received props

kfiroo:
@jeromegrosse Great work!
Regarding your questions, maybe @FaridSafi would have an opinion :)

* I would say put the ticks on the right so the `time` won't move when the status changes
* I think only sent messages should show ticks for now, if users will request we can easily add it to the others

jeromegrosse:
@kfiroo @FaridSafi 
I've just pushed the changes that you suggested:
* The ticks are now display on the right-hand side of the time.
* I fixed a styling problem that moved the time on the left of the message. It is now display on the right of the messages as it currently is in the master branch. Maybe now the ticks would be better on the left-hand side of the time?
* There is a check that prevent the ticks from being displayed for the received messages.
* I moved up the if statement as you suggested to allow an easier overriding of the function.
* I modified the example so as to include the ticks feature.

src/Bubble.js
```diff
@@ -52,6 +53,25 @@ export default class Bubble extends React.Component {
     return null;
   }
 
+  renderTicks() {
+    if (this.props.renderTicks) {
+        return this.props.renderTicks(currentMessage);
+    }
+
+    const {currentMessage} = this.props;
```
kfiroo: This line should be above the if statement - you are using `currentMessage` and it is not defined

kfiroo:
@jeromegrosse Yeah, I would move them to the left now :) sorry!
That + the comment I left inline should have this feature closed I think.
@FaridSafi what do you think?

src/Bubble.js
jeromegrosse: Stupid overlook from my part. Nice catch!

jeromegrosse:
I've push the change with the constant declaration that is declared after its first use if the function is overridden.

Regarding the ticks on the left, I've encountered a problem: The `Time` component has a `marginLeft` set to 10 which, I believe, is not that easy to override. The best result I could get with the ticks on the left is shown in the screenshot below:
photo_2016-09-11_00-01-56

jeromegrosse:
Have you got time to review it @kfiroo @FaridSafi  ?

kfiroo:
@jeromegrosse @FaridSafi looks good to me :)

jeromegrosse:
@kfiroo @FaridSafi Just a little update to be able to customize how ticks are displayed (mainly overwriting the colour) by simple passing `tickStyle` to `GiftedChat`.

benoist:
Any update on this?

kfiroo:
@dgdavid @cooperka @scarmuega would love your opinion on this

cooperka:
Code LGTM, haven't tested on a device though. Nice work.

dgdavid:
The code also looks good to me, but I've some doubts about merging it directly or improve it a little before.

I'll try to explain myself. As @cooperka said, this is a `gifted` chat component and this feature should be out-of-the-box. We agree on this, I think. But, Why on the left-side? Why with plain text and not using icons? Why there are not a `sending` status to give feedback to the user? I now that all these questions are out of the component's scope because each application should implement its own logic for each particular behavior. However, we can do little changes to make this easier.

I would split `Bubble` into three parts: _Header_, _Content_, and _Footer_. Then use the `renderTime` and `renderTick` methods into  `renderBubbleFooter` or similar. In my opinion, this way gives a more clean logic and allows to override the entire footer at once without bloating `GiftedChat` with a lot of `renderXXX` props, since to change the order of ticks and time you could do something such as

```javascript
...
renderBoobleFooter={ function() {
  renderTime();
  renderTicks();
} }
...
```

or change the entire footer passing a completely different logic.

As a more detailed example, below you have some of code that I'm using currently in my project:

```javascript
...

  messageFooter(timeProps) {
    let icon = null
    const { currentMessage, position } = timeProps
    const time = moment.tz(currentMessage.createdAt, timezone).format('LT')
    const text = <Text style={styles[position].text}>{ time }</Text>

    // TODO: improve the icon logic
    if (showStatusIcon(currentMessage.status)) {
      icon =
          <Icon
            name={statusIcons[currentMessage.status]}
            style={[styles[position].text, styles[position].icon]}
          />
    }

    return (
      <View style={styles[position].container}>
        { text }
        {
          currentMessage.status === 'FAILED'
            ? <TouchableOpacity onPress={() => this.onSend([currentMessage])}>{icon}</TouchableOpacity>
            : icon
        }
      </View>
    )
  }

  render() {
    return (
      <View style={{ flex: 1 }}>
        <GiftedChat
          style={styles.mainContainer}
          messages={this.props.messages && this.props.messages.map(m => unflatten(m))}
          onSend={this.onSend}
          messageIdGenerator={uuid.v4}
          renderTime={this.messageFooter}
          user={{
            _id: this.props.currentUser.id,
          }}
        />
      </View>
    );
  }
}
...
```

I hope that you understand what I trying to say.

cooperka:
@dgdavid: I agree with all points, though I think most of that can be held off for a future PR. I'm always in favor of merging early with an [MVP](http://blog.crisp.se/2016/01/25/henrikkniberg/making-sense-of-mvp) and iterating from there. This PR is a fairly simple change that allows the possibility of showing send-receipts, and your suggestions seem to me like a larger structural change to the whole component (possibly breaking). I agree it should be done as you suggest, but perhaps in a separate PR after this one?

In terms of UX, WhatsApp (just as an example) uses icons on the right of the timestamp. Telegram is the same way. Below is a nice description of what each icon means on WhatsApp, showing how complex this could eventually be. Again, I don't think we need this now, but it's nice to see an example:

whatsapp-tick-marks-meanings

TL;DR: :shipit:

dgdavid:
@cooperka I also agree with you. I proposed to do it in this PR by the relation with the `renderTime` prop/method.  But you're right, possibly it would be a big change across the whole component.

jeromegrosse:
@dgdavid @cooperka 
The ideas that you are proposing are definitely worth thinking about. When I first started this PR, the scope of what I was aiming for was way less than what you are suggesting. However, I'd be glad to continue this PR and/or help with a subsequent PR to help implement these new features.

kfiroo:
@jeromegrosse @cooperka @dgdavid Hey guys
So glad to see everybody's on board with improving this project!
I think this PR is pretty good on it's own, it provides a minimal functionality with minimal changes and as a "feature PR" it is pretty solid.

However, I would like to open a new issue(s?) for discussing the future of `GiftedChat`, where we should talk about:
* Code style (eslint)
* Technical issues (tests / `GiftedChat` API / `renderXXX` inconsistencies / etc)
* Missing features (ticks / image zoom / others?)

Not so sure if we need a single issue or an issue for each topic but I guess we'll find out soon enough :)

cooperka:
Friendly ping @kfiroo

guns2410:
Please push to npm as well

Maxwell2022:
Please add `docs` as well

vvavepacket:
Can we add docs of this in front page?

flashimxd:
This feature is already implemented? I couldn't find in the documentation, and I really need that..

Thank you guys!

Faolain:
Hey everyone any thoughts on the implementation here https://github.com/FaridSafi/react-native-gifted-chat/issues/644 for doing essentially the same thing but outside of the bubble?

Usamaliaquat123:
any update about seen messages and unseen message? and how to implement the isTyping option?

hindsricardo:
?? soooo, is seen implemented or not? If so, where is the documentation?

mkhoussid:
> any update about seen messages and unseen message? and how to implement the isTyping option?

Are you using Socket IO? Either way, use the `onInputTextChanged` prop.

Thanujabadabagni:
hi am unable to bind messages to giftedchat messages using global list.Iin contructor i declared like global.messagesList=[];

<GiftedChat
        messages={global.messagesList}
      //  onSend={messages => this.onSend(messages)}
        user={{
          _id:this.state.LoginId,
        }}
      />
please suggest  how to use global list to giftedchat messages...

rusakovic:
> > Any update about this how we can implement a tick when delever and seen message
> 
> To message object add
> 
> ```
>             sent: true,
>             received: true,
>             pending: false,
> ```

Please, add it into the Readme file

nwdeer:
So, is it possible to indicate that the message was not only delivered but also read?

Issue #181, opened 2016-08-09
News shows that starts from the top


289420843:
A small amount of message height do not meet the container height. Now is displayed at the bottom. Hope to start from the top.How to do it?

longsangstan:
Are you saying that when there are only a few messages, they are not sticking to the top?
I have the same question as well.

289420843:
@longsangstan That's right.

longsangstan:
Why do you close it? You have fixed the issue?

289420843:
Open this issue, the gifted-chat to solve the issue.

kfiroo:
@289420843 @longsangstan This is indeed an issue with `react-native-invertible-scroll-view` which is still open, hopefully a solid fix will come up soon :)

I tried to get it to work with `gifted-chat` but it messed up a few other things so I left it for now..

satyagupta:
Do array.reverse and then load that as messages. 

And When adding a new message object using GiftedChat.append function to add new message object.

hegelstad:
If somebody is stuck with `Inverted` prop not working, remember to refresh app/remount component.

PR #231, opened 2016-09-09
Align messages from the top

maxs15:
Here is the fix for #181 using the solution provided by ide (https://github.com/exponentjs/react-native-invertible-scroll-view/issues/11)


maxs15:
Damn...the fix doesn't work when the list is getting longer, breaks the scroll :/

PR #240, opened 2016-09-20
Set send setTimeout to 0

bsiddiqui:
Currently messages can be cleared if you're in the middle of typing the
next message when setTimeout resets the toolbar


bsiddiqui:
@alanhof

PR #246, opened 2016-09-26
Missing overwrite method for the renderAvatar

justinpark:
The message component looked missing the renderAvatar overwrite method.


kfiroo:
@justinpark Hey! 
This method is actually in [Avatar.js#L12](https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/Avatar.js#L12).
I have to say I like it better when the `renderXXX` method sits outside the `XXX` component.
If your PR could include removing the other override it would be better.

Thanks

PR #251, opened 2016-09-29
Add callback on Press message's image

Fantasim:
Add callback onPressImage which take in parameter image's url, to show image in modal (for example), name callback function : onPressImage


delvedor:
@FaridSafi this would be very useful! :)
Are you planning to merge this once the conflicts are resolved?

PR #252, opened 2016-09-30
Updated imageStyle propType to match usage

iggyfisk:


cooperka:
Hi @iggyfisk, looks like you have an extra commit in here that goes beyond the scope of this PR. Please remove it and I'll be happy to take another look.

PR #254, opened 2016-10-04
Expose GiftedAvatar

jeromegrosse:
Export  `GiftedAvatar` in `GiftedChat.js` so as to be able to use `GiftedAvatar` without all the logic of `Avatar`.


jeromegrosse:
So, what's up? @kfiroo

Issue #256, opened 2016-10-05
Style variable name conflicting in Send and MessageText


jeromegrosse:
#### Issue Description

Conflict over the property: `TextStyle` as it modify both the text within the Bubble and the Text within the Send button.
#### Steps to Reproduce / Code Snippets

Pass the property `TextStyle` to GiftedChat. If it is passed so as to validate the properties of `MessageText` (i.e.: With a `left` and `right`), the Send button won't be correctly displayed. On the other case, `TextStyle` is passed so as to modify the style of the Send button, it won't validate and `MessageText` and ReactNative will throw an error.
#### Expected Results

Have two different style name for MessageText and Send so as to not have a conflict and be able to modify them independently.
#### Additional Information
- React Native version: 0.34.1
- react-native-gifted-chat version: 0.0.10
- Platform(s) (iOS, Android, or both?): both

kfiroo:
@jeromegrosse I would suggest you pass `renderSend` and render the `Send` component with the props you need.

jpamarohorta:
A similar problem affects `containerStyle` from Send and  `containerStyle` from InputToolbar. 

Maybe we could create a naming rule for style props like <ComponentNameStyleName>. For this example the props would change to `sendContainerStyle` and `inputToolbarContainerStyle`.

Any issues with this change?

lukasbalaz:
@kfiroo do you mean that I should first pass `containerStyle` prop to `InputToolbar` and then override `renderSend` function just to get rid of it ? Doesn't look very clean to me, I think it would be better not to pass custom style props to `render{Component}` methods at all because:
1. It is not clear that this is how it works. I just spent an hour debugging why is the the Send button styled weirdly when I change the InputToolbar style.
2. You almost never want to pass the same `containerStyle` prop to both parent and child, and if you really do, then you can override the `render` method and not the other way around

What do you think ?

Nabellaleen:
The bug is still there, this issue should not be closed.

Many components are receiving "containerStyle" prop, but with various typing (PropTypes.shape left/right in Avatar or Message, ViewPropStyle in InputToolbar or Day, for examples)

This should be fixed :)

PR #257, opened 2016-10-05
Send: Change textStyle to sendTextStyle to avoid collisions with MessageText

jeromegrosse:
Related to issue #256


jeromegrosse:
So, what's up? @kfiroo

kfiroo:
@jeromegrosse I can accept this PR, but it might break some users who are relying on it.
Wouldn't it be easier for you to use `renderSend`?

PR #259, opened 2016-10-09
Allow to change how new messages' ids are generated

dgdavid:
Hi @FaridSafi!

First of all, thanks to you and the collaborators for the work in this component.

 I'm testing it in a project in which the team wants to use UUID for new messages' ids. So, I made a small change adding a new prop to `GiftedChat` called `messageIdGenerator` that receive a function.

I also updated the example app to add this prop using [`react-native-uuid`](https://github.com/eugenehp/react-native-uuid).

Regards.


kfiroo:
@dgdavid I think you should use [node-uuid](https://github.com/broofa/node-uuid), looks like `react-native-uuid` is simply a copy of that one

kfiroo:
@dgdavid Oh, it's only in the example, I think we can use it as the default id-generator.
btw, did you make sure it doesn't have a performance penalty?

dgdavid:
@kfiroo you're right, I didn't realize. Sure is better to use `node-uuid`.

About performance, actually I didn't make any test yet because I'm only playing with all necessary pieces at this moment.

Anyway, I believe that is a good idea keep the prop to allow users to change the function if they need it.

kfiroo:
@dgdavid Yeah, I agree, the prop should stay.
Just set the default to use 'node-uuid' and not the current function

dgdavid:
Ok, I changed it and remove modifications on the example app.

But remember that, as I said above, I didn't test how this affect to performance.

dgdavid:
Hey @kfiroo 

Do you plan to merge this?

kfiroo:
@dgdavid Sure!

PR #261, opened 2016-10-11
feature: allow to pass rowHasChanged as prop

sondremare:


kfiroo:
@sondremare Just out of curiosity, why would you need this? :)

PR #265, opened 2016-10-20
Exposes currentMessage to props passed onLongPress

jesperlndk:
This makes it possible to actually do stuff with the currentMessage that one is long-pressing.


jesperlndk:
@FaridSafi

dgdavid:
@xcarpentier, 

The change introduced by this PR also was added by you in the commit 28cb4675285c80b70a0f042bddde0e43ece47a0a, if I am not mistaken.

So, you can close it.

PR #274, opened 2016-10-31
Fix: Incorrect switch statement for Call & Text

guns2410:
Fixed the incorrect switch statement for Call and Text on phone number pressed!


guns2410:
Closing in favour of #418

PR #242, opened 2016-09-20
Add onTopReached callback for auto loading earlier messages

ramilushev:

PR #275, opened 2016-11-03
On top reached

arturictus:
Usefull to load previous messages when user scrolls to the top

### DIFF
--- a/src/MessageContainer.js
+++ b/src/MessageContainer.js
@@ -135,10 +135,20 @@ export default class MessageContainer extends React.Component {
     );
   }
 
+  endReachedProps(){
+    let condProps = {}
+    if (this.props.onTopReached){ condProps.onEndReached = this.props.onTopReached }
+    if (this.props.onTopReachedThreshold){
+      condProps.onEndReachedThreshold = this.props.onTopReachedThreshold
+    }
+    return condProps
+  }
+
   render() {
     return (
       <View ref='container' style={{flex:1}}>
         <ListView
+          {...this.endReachedProps()}
           enableEmptySections={true}
           keyboardShouldPersistTaps={true}
           automaticallyAdjustContentInsets={false}
--- a/README.md
+++ b/README.md
@@ -118,6 +118,8 @@ See [example/App.js](example/App.js)
 - **`renderSend`** _(Function)_ - render the send button
 - **`renderAccessory`** _(Function)_ - renders a second line of actions below the message composer
 - **`bottomOffset`** _(Integer)_ - distance of the chat from the bottom of the screen, useful if you display a tab bar
+- **`onTopReached`** _(Function)_ - function to call when scroll top is reached
+- **`onTopReachedThreshold`** _(Integer)_ - distance to the top to trigger onTopReached
 
 
 ## Features


src/MessageContainer.js
```diff
@@ -135,10 +135,20 @@ export default class MessageContainer extends React.Component {
     );
   }
 
+  endReachedProps(){
+    let condProps = {}
+    if (this.props.onTopReached){ condProps.onEndReached = this.props.onTopReached }
```
dgdavid: Sorry if I'm a little bit confused but, Are you using `onEndReached` to know if you're on top of list?

Maybe you wanted to use [`onScroll`](https://facebook.github.io/react-native/docs/scrollview.html#onscroll) property?

src/MessageContainer.js
```diff
@@ -135,10 +135,20 @@ export default class MessageContainer extends React.Component {
...
+      condProps.onEndReachedThreshold = this.props.onTopReachedThreshold
+    }
+    return condProps
+  }
+
   render() {
     return (
       <View ref='container' style={{flex:1}}>
         <ListView
+          {...this.endReachedProps()}
```
kfiroo: @arturictus Why not pass the props directly to the component?
What is the point of wrapping it in a function that returns the object?

src/MessageContainer.js
kfiroo: @dgdavid I think this is the right way, the underlaying scroll component is an `InvertibleScrollView` so the top is the end.
@arturictus Am I getting this right?

dgdavid: @kfiroo, @arturictus Ups, sorry :sweat_smile: :sweat_smile:. I didn't realize that messages `ListView` is using a [`InvertibleScrollView`](https://github.com/exponentjs/react-native-invertible-scroll-view). I thought that it was a `ScrollView`

arturictus: Hi @dgdavid and @kfiroo ,
The `InvertibleScrollView` receives `onEndReached` as `ScrollView`,  but the Y axis is inverted so it triggers `endReached` when is at the top.
On scroll is not as reliable as `end/topReached` reached is only triggered once onScroll is more unpredictable.

arturictus: What I'm thinking is that maybe this pull request should be done at the `InvertibleScrollView` so this component will benefit from it as all the others using `InvertibleScrollView`.
what do you think @dgdavid and @kfiroo ?

dgdavid: In my opinion it seems a good idea.

Anyway, in case that your PR will be accepted there, you need to make a new one here to allow pass the function and threshold through `GiftedChat` (maybe arround [here](https://github.com/FaridSafi/react-native-gifted-chat/blob/91312c16f041bd78c61d98d0cbdf3f539a7f0f24/src/GiftedChat.js#L68)?) and add some documentation about the `MessageContainer` component uses a [`InvertibleScrollView`](https://github.com/exponentjs/react-native-invertible-scroll-view), explaining that actually _onTopReached_ and _onTopReachedThreshold_ are only alias of _onEndReached_ and _onEndReachedThreshold_ in order to improve the comprehension.

Isn't it right?

src/MessageContainer.js
dgdavid: @kfiroo I think he was trying avoid to pass empty props. One solution, instead use an additional function, could be to add the defaults values.

cooperka:
Closing in favor of earlier https://github.com/FaridSafi/react-native-gifted-chat/pull/242, which now has one of your commits cherry-picked in. Thanks for the PR!

PR #280, opened 2016-11-14
Support for orientation changes.

psrustik:
Turned on onLayout handler for all platforms.

Android only: on-screen keyboard resized main container and has own height.
https://developer.android.com/training/keyboard-input/visibility.html
So for calculate the messages container height need ignore keyboard height.

Current problem after orientation changes:
current_1
current_2

With our fix: 
after_fix_1
after_fix_2

### DIFF
--- a/src/GiftedChat.js
+++ b/src/GiftedChat.js
@@ -154,7 +154,14 @@ class GiftedChat extends React.Component {
   }
 
   getKeyboardHeight() {
-    return this._keyboardHeight;
+    if (Platform.OS === 'android') {
+      // For android: on-screen keyboard resized main container and has own height.
+      // @see https://developer.android.com/training/keyboard-input/visibility.html
+      // So for calculate the messages container height ignore keyboard height.
+      return 0;
+    } else {
+      return this._keyboardHeight;
+    }
   }
 
   setBottomOffset(value) {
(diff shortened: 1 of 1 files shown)


psrustik:
Current problem after orientation changes (ios):
---
before_1
---
before_2

With our fix:
---
after_fix_1
---
after_fix_2

cooperka:
Unfortunately I think this change no longer works, even after rebase. It causes the keyboard to cover the TextInput on Android. @psrustik any thoughts on how to fix it?

cooperka:
Ah, it was my mistake. I didn't realize you had changed `&&` to `||` and it was lost while resolving rebase conflicts. Tested and verified on latest master, both Android and iOS :+1:

@psrustik or @kfiroo you can cherry-pick my rebased work here to fix the merge conflicts if you want to: https://github.com/cribspot/react-native-gifted-chat/commit/09ab4e90544c54103711281f31cba2a94ef28c27

It would be great to have this PR merged!

kfiroo:
@psrustik Thanks for the PR, if you could resolve the conflicts I would gladly merge it :)
sorry for the delay, I was AFK for a month

cooperka:
This is an existing issue on Android, but this PR makes it occur on iOS too: https://github.com/FaridSafi/react-native-gifted-chat/issues/413. Because of this, we should probably resolve the whole issue in this PR before merging.

cooperka:
This PR can be closed in favor of https://github.com/FaridSafi/react-native-gifted-chat/pull/425 (which has this commit rebased on latest master).

PR #287, opened 2016-11-19
Adding Footer for ToolBar. This footer can show custom view like facebook or snapchat show in keyboard area.

hetang-shah:
facebook-keyboard-bar


asessa:
Can't you do this by using renderAccessory ?

PR #297, opened 2016-12-05
Added "alwaysRenderAvatar" prop to "Avatar.js"

eomine:
Currently, the `Avatar` component hides the avatar of a message if it's from the same day and from the same user of the previous message.

This prop allows users to disable this behavior, so that the avatar is rendered on all messages.

The default value is `false`, which keeps the current behavior.

In order to render the avatar on all messages, one should set the `alwaysRenderAvatar` prop to `true` when placing the `GiftedChat` component.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ See [example/App.js](example/App.js)
 - **`renderSend`** _(Function)_ - render the send button
 - **`renderAccessory`** _(Function)_ - renders a second line of actions below the message composer
 - **`bottomOffset`** _(Integer)_ - distance of the chat from the bottom of the screen, useful if you display a tab bar
-- **`alwaysRenderAvatar`** _(Bool)_ - always render the message avatar, even when consecutive messages are from same day and same user
+- **`alwaysRenderAvatar`** _(Bool)_ - always render the message avatar, even when consecutive messages are from same day and same user. The default value is `false`.
 
 
 ## Features
--- a/README.md
+++ b/README.md
@@ -118,6 +118,7 @@ See [example/App.js](example/App.js)
 - **`renderSend`** _(Function)_ - render the send button
 - **`renderAccessory`** _(Function)_ - renders a second line of actions below the message composer
 - **`bottomOffset`** _(Integer)_ - distance of the chat from the bottom of the screen, useful if you display a tab bar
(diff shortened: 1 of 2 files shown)


eomine:
@dgdavid Thanks!

Changed the prop to a shorter name. Its value must be set to `true` now, which seems more straightforward.

Updated the Readme.

src/Avatar.js
```diff
@@ -63,6 +67,7 @@ const styles = {
 };
 
 Avatar.defaultProps = {
+  alwaysRenderAvatar: false,
```
dgdavid: Cool! Looks better now. I was thinking a shorter name but no idea came to my mind ;)

README.md
```diff
@@ -118,6 +118,7 @@ See [example/App.js](example/App.js)
 - **`renderSend`** _(Function)_ - render the send button
 - **`renderAccessory`** _(Function)_ - renders a second line of actions below the message composer
 - **`bottomOffset`** _(Integer)_ - distance of the chat from the bottom of the screen, useful if you display a tab bar
+- **`alwaysRenderAvatar`** _(Bool)_ - always render the message avatar, even when consecutive messages are from same day and same user
```
dgdavid: It would be interesting to indicate the default value.

eomine: Ok, updated the readme with the default value.

prithsharma:
@dgdavid since #292 is merged now, is there anything else blocking this PR right now?
Thanks for the update, in advance.

kfiroo:
@prithsharma Looking at the code I'm not sure it will help you after all :(
This only overrides the `avatar visibility` for **received** message.

dgdavid:
@prithsharma It would be nice to use the `utils` functions after import them instead rely on `props` in order to follow the way introduced by @scarmuega in #292.

For example, replacing `this.props.isSameUser` with `isSameUser` after `import { isSameUser } from './utils'`.

But these days I'm a little out of the loop to think/discuss the viability of this PR, so once again we need the @kfiroo support.

zachrnolan:
Hey @eomine @dgdavid @kfiroo any updates on this? Would love to see this feature merged.

dgdavid:
@xcarpentier 

Since you have now a good knowledte of project, could please take a look to this PR and give your opinion about what to do in order to fix and merge or reject it?

Thanks!

hironarita:
Will this eventually be merged?

apparition47:
@hironarita #729 was merged

PR #301, opened 2016-12-08
Add prop onInputTextChanged on GiftedChat API and help fix #286

### LINKED ISSUE #286: Strange behavior with onChange in overridden Composer for typing indicator for that user
#### Issue Description

By using the`renderComposer` to access the onChange prop it will disable inherited props like `renderSend` and `this.props.text` and the growing height of the TextInput.

```javascript
  renderComposer(props) {
    return (
      <Composer {...props} onChange={() => this.handleChangeText()} />
    )
  }
```

#### Expected Results

Composer remains intact and onChange sends back event information to be able to fire off that the user is typing.

#### Additional Information

* React Native version: 0.36.0
* react-native-gifted-chat version: 0.0.6
* Platform(s) (iOS, Android, or both?): iOS

jayshah123:
This introduces a new prop : onInputTextChanged on GiftedChat API,
which is a callback function that will give whatever text is present currently in Composer back to user.
Composer has onChange prop but it is kind of consumed by onType function in GiftedChat.js.
This will also help fix #286, where you see GiftedChat.js personally uses onChange,
and the behaviour such as height calculation would have to be performed by library user (which right now happens in onType in GiftedChat.js) 
This prop gives users a simpler interface to track change in input text, with all other features intact.
Kindly guide with suggestions/comments wherever feels necessary.
Thank you.

### DIFF
--- a/src/Message.js
+++ b/src/Message.js
@@ -82,7 +82,7 @@ export default class Message extends React.Component {
       <View>
         {this.renderDay()}
         <View style={[styles[this.props.position].container, {
-          marginBottom: this.isSameUser(this.props.currentMessage, this.props.nextMessage) ? 2 : 10,
+          marginBottom: this.isSameUser(this.props.currentMessage, this.props.nextMessage) ? this.props.marginSameUser : this.props.marginDifferentUser,
         }, this.props.containerStyle[this.props.position]]}>
           {this.props.position === 'left' ? this.renderAvatar() : null}
           {this.renderBubble()}
@@ -119,6 +119,8 @@ Message.defaultProps = {
   renderBubble: null,
   renderDay: null,
   position: 'left',
+  marginSameUser: 6,
+  marginDifferentUser: 10,
   currentMessage: {},
   nextMessage: {},
   previousMessage: {},
@@ -130,6 +132,8 @@ Message.propTypes = {
   renderAvatar: React.PropTypes.func,
   renderBubble: React.PropTypes.func,
   renderDay: React.PropTypes.func,
+  marginSameUser: React.PropTypes.number,
+  marginDifferentUser: React.PropTypes.number,
   position: React.PropTypes.oneOf(['left', 'right']),
   currentMessage: React.PropTypes.object,
   nextMessage: React.PropTypes.object,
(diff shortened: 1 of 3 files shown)


src/Message.js
```diff
@@ -82,7 +82,7 @@ export default class Message extends React.Component {
       <View>
         {this.renderDay()}
         <View style={[styles[this.props.position].container, {
-          marginBottom: this.isSameUser(this.props.currentMessage, this.props.nextMessage) ? 2 : 10,
+          marginBottom: this.isSameUser(this.props.currentMessage, this.props.nextMessage) ? this.props.marginSameUser : this.props.marginDifferentUser,
```
dgdavid: Hi @jayshah123 

I think that this is not related to the original PR. In my opinion, is better to split this changes in two different PR.

src/Message.js
```diff
@@ -119,6 +119,8 @@ Message.defaultProps = {
   renderBubble: null,
   renderDay: null,
   position: 'left',
+  marginSameUser: 6,
```
dgdavid: The original value was `2`. Just for curiosity, Why do you want to change it?

src/Message.js
kfiroo: @jayshah123  This is how github works, you stated that the master branch of your fork is the source of the PR so every change in it will be included, you could use a feature branch to achieve what you want

jayshah123: Ok I think I will revert the second commit and that should remove it?

kfiroo: Not sure :)
Give it a try

jayshah123: Rewriting history is Ok? I will close PR, rewrite history and raise a fresh PR(with single commit).

dgdavid: @jayshah123 Don't worry.

You only need to create a new branch in your fork, make the commit on it and create a new PR in the same way you created the previous one.

My recommendation is that this new branch will be created from `master` (in this case).

The Github documentation is better explained that my comment. Here you have some links. I hope it'll be useful for you.

https://guides.github.com/introduction/flow/
https://help.github.com/articles/creating-a-pull-request/

jayshah123:
I'm a novice github user,
messed up this PR by adding a unnecessary commit on master branch.
Will raise a proper PR.

src/Message.js
jayshah123: Aha Got it thanks!

PR #304, opened 2016-12-12
Add lightbox for enlarging chat images

cooperka:
Images sent in messages will be enlarged when you tap them.

Also adds 2 new props: `imageProps` and `lightboxProps`. These will be passed through to the `<Image>` and `<Lightbox>`, respectively. For example, you can now specify an image [`onLoad`](https://facebook.github.io/react-native/docs/image.html#onload) callback.

Tested and verified on iOS and Android.

---

Note: Uses a commit hash instead of version number in order to fix a warning on Android that's currently unpublished.

You can read more about react-native-lightbox [here](https://github.com/oblador/react-native-lightbox) :tropical_drink:

### DIFF
--- a/src/MessageImage.js
+++ b/src/MessageImage.js
@@ -20,6 +20,7 @@ export default class MessageImage extends React.Component {
           }}
         >
           <Image
+            {...this.props.imageProps}
             style={[styles.image, this.props.imageStyle]}
             source={{uri: this.props.currentMessage.image}}
           />
@@ -50,6 +51,7 @@ MessageImage.defaultProps = {
   },
   containerStyle: {},
   imageStyle: {},
+  imageProps: {},
   navigator: null,
 };
 
@@ -57,5 +59,6 @@ MessageImage.propTypes = {
   currentMessage: React.PropTypes.object,
   containerStyle: View.propTypes.style,
   imageStyle: Image.propTypes.style,
+  imageProps: React.PropTypes.object,
   navigator: React.PropTypes.object,
 };
--- a/src/MessageImage.js
+++ b/src/MessageImage.js
@@ -14,10 +14,10 @@ export default class MessageImage extends React.Component {
     return (
       <View style={[styles.container, this.props.containerStyle]}>
         <Lightbox
-          navigator={this.props.navigator}
           activeProps={{
             style: [styles.imageActive, { width, height }],
           }}
+          {...this.props.lightboxProps}
         >
           <Image
             {...this.props.imageProps}
@@ -52,7 +52,7 @@ MessageImage.defaultProps = {
   containerStyle: {},
   imageStyle: {},
   imageProps: {},
-  navigator: null,
+  lightboxProps: {},
 };
 
 MessageImage.propTypes = {
@@ -60,5 +60,5 @@ MessageImage.propTypes = {
   containerStyle: View.propTypes.style,
   imageStyle: Image.propTypes.style,
   imageProps: React.PropTypes.object,
-  navigator: React.PropTypes.object,
+  lightboxProps: React.PropTypes.object,
 };
--- a/src/MessageImage.js
+++ b/src/MessageImage.js
@@ -3,16 +3,27 @@ import {
   Image,
   StyleSheet,
   View,
+  Dimensions,
 } from 'react-native';
+import Lightbox from 'react-native-lightbox';
 
 export default class MessageImage extends React.Component {
   render() {
+    const { width, height } = Dimensions.get('window');
+
     return (
       <View style={[styles.container, this.props.containerStyle]}>
-        <Image
-          style={[styles.image, this.props.imageStyle]}
-          source={{uri: this.props.currentMessage.image}}
-        />
+        <Lightbox
+          navigator={this.props.navigator}
+          activeProps={{
+            style: [styles.imageActive, { width, height }],
+          }}
+        >
+          <Image
+            style={[styles.image, this.props.imageStyle]}
+            source={{uri: this.props.currentMessage.image}}
+          />
+        </Lightbox>
       </View>
     );
   }
@@ -27,6 +38,9 @@ const styles = StyleSheet.create({
     borderRadius: 13,
     margin: 3,
     resizeMode: 'cover',
+  },
+  imageActive: {
+    resizeMode: 'contain',
   },
 });
 
--- a/src/MessageImage.js
+++ b/src/MessageImage.js
@@ -50,10 +50,12 @@ MessageImage.defaultProps = {
   },
   containerStyle: {},
   imageStyle: {},
+  navigator: null,
 };
 
 MessageImage.propTypes = {
   currentMessage: React.PropTypes.object,
   containerStyle: View.propTypes.style,
   imageStyle: Image.propTypes.style,
+  navigator: React.PropTypes.object,
 };
--- a/README.md
+++ b/README.md
@@ -108,6 +108,7 @@ See [example/App.js](example/App.js)
 - **`renderMessage`** _(Function)_ - render the message container
 - **`renderMessageText`** _(Function)_ - render the message text
 - **`renderMessageImage`** _(Function)_ - render the message image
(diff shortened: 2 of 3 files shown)


src/MessageImage.js
```diff
@@ -3,16 +3,27 @@ import {
...
+    const { width, height } = Dimensions.get('window');
+
     return (
       <View style={[styles.container, this.props.containerStyle]}>
-        <Image
-          style={[styles.image, this.props.imageStyle]}
-          source={{uri: this.props.currentMessage.image}}
-        />
+        <Lightbox
+          navigator={this.props.navigator}
```
kfiroo: @cooperka I don't think you can assume `this.props.navigator` exists.
Even if it does, not everyone uses react-natives `Navigator` which is what this package is assuming.

kfiroo:
@cooperka Hey! Thanks for the PR :)
I'm not sure that this fits the scope of this component, plus, it can be super easily achieved with overriding `renderMessage` / `renderMessageImage`.

I would love to hear what others have to say though

src/MessageImage.js
cooperka: If `navigator` is null, the lightbox still renders just fine :) In fact, I don't even use it myself. It's [recommended by lightbox](https://github.com/oblador/react-native-lightbox#usage), but not necessary by any means.

cooperka:
Thanks @kfiroo! Honestly to me it seems like an oversight that this feature **doesn't** already exist, as everyone I've shown the chat interface to has tried to tap my mock image to see it larger and been disappointed when nothing happens. We could definitely let people implement this on a per-app basis instead, but that just seems a little redundant to me. This chat is supposed to be "gifted" 😉. I'm also interested to see what other people have to say.

src/MessageImage.js
cooperka: I just added it to `propTypes` where it defaults to `null`. Hopefully this helps.

dgdavid:
Guys, at first I agree with @kfiroo's doubts about the component scope. But honestly we need to admit that @cooperka have good, strong arguments. So, I keep my thumb up for this feature.

kfiroo:
@cooperka @dgdavid Yeah, I hear you. I'm starting to turn too :)

Few questions though, 
1. Is the `Lightbox`ed image zoomable? I feel like this is a must if we are going with this.
2. We need to be able to use a custom `Image` component (like `ResponsiveImage` / `CachedImage` etc.)
3. Do we want to be able to customize the `Lightbox` view (add some info / actions)?

cooperka:
@kfiroo it looks like there are a [ton of new features coming as part of a major refactor](https://github.com/oblador/react-native-lightbox/issues/8#issuecomment-160352712), and I'm not quite sure but maybe it will even be able to use its own instance of `Lightbox.Navigator` if you don't pass one in yourself? The GIF on that linked comment shows off the new stuff. Unfortunately progress seems to have halted on the refactor... I added a comment to inquire.

**Before** the refactor (currently):

1. Lightbox is not zoomable, but the purpose of Lightbox is to zoom the image that was sent in chat. I feel like Lightbox is a must if we are going to allow images sent in chat, even if it lacks some features in the short-term
2. Arbitrary custom content should already work, Lightbox seems to be able to zoom any child content, not just images (see the GIF in their README)
3. That would be great! See below.

**After** the refactor:

1. Yes
2. Same answer
3. It looks like this is made much easier by the refactor

TL;DR: I think we should go ahead and merge this PR, and if needed we/I can help contribute to Lightbox to bring it up to par with where we want it in the long-term.

kfiroo:
@cooperka take a look at this fork https://github.com/shoutem/react-native-lightbox

cooperka:
@kfiroo Nice! Want me to use that one here instead?

kfiroo:
@cooperka Not sure actually, we should use the one that works better :)
I'm also fiddling with "zoom views" now, I'll let you know if I have anything interesting.
Here's another interesting project https://github.com/ldn0x7dc/react-native-view-transformer
Which has a child projects https://github.com/ldn0x7dc/react-native-transformable-image and https://github.com/ldn0x7dc/react-native-gallery

cooperka:
Interesting, let me know how it turns out! Lightbox does seem by far the most popular of the options, and intends to cover the gallery use-case with the refactor, but if something else works better I'm all for it.

kfiroo:
I wouldn't hold my breath waiting for that refactor to land.. repo hasn't changed in a year or so

cooperka:
Definitely agree. I asked about the status in that thread (https://github.com/oblador/react-native-lightbox/issues/8#issuecomment-267025835), and at worst I'll just fork the repo in order to merge a few of the nice PRs there. Should work fine either way.

cooperka:
@kfiroo there's an update here: https://github.com/oblador/react-native-lightbox/issues/8. Honestly we don't need all the bells and whistles from the proposed refactor of lightbox, but there are several PRs that would be nice to have merged. Shall we fork it?

kfiroo:
@cooperka I didn't like how it works, but I guess it might be a good enough, out-of-the-box solution for many users, so I we can merge it or try with the `navigator-refactor`, what ever you think is better.
I'm only missing a way to disable it, but I guess you could override `renderMessageImage` and render what ever you want.

In my app, I ended up writing something myself, I can try to extract it to its own repo and see if we like it better.

cooperka:
I agree that lightbox is not ideal... if you can extract your logic into something simple and usable I'm all for it! How long do you expect that to take?

kfiroo:
@cooperka I'll try to do it this weekend

cavneb:
+1

cooperka:
@kfiroo how goes your extraction?

I updated the props a bit on this PR, I think it's better and more generic after the changes. Also allows passing image props like `onLoad` now.

kfiroo:
@cooperka I couldn't find time to do it, I think we can merge your PR for now and get to it later if I ever get to it :)
Can I merge this? Are we good on both android and ios?

cooperka:
Yes, good to merge! I tested it just now after adding the new prop.

booboothefool:
@cooperka Does this work out of the box, or is there something I have to configure to get the lightbox to pop up? I simply tried updating to the latest version of react-native-gifted-chat and clicked on an image, but nothing happened. Sorry for the stupid question. 😛

cooperka:
@booboothefool yeah it should work out of the box, no setup needed! You can try getting it to work in the [example app](https://github.com/FaridSafi/react-native-gifted-chat/tree/master/example), and if it's still not working just send us a link to your code and we can take a look.

Might be silly, but also make sure you've run `npm install` and reloaded your app after updating your version of the library.

Also, if you're passing in a custom component e.g. `renderBubble` or `renderMessageImage` then of course you won't get the new changes unless you update the component yourself.

booboothefool:
@cooperka Well I got everything hooked up, but do you notice any performance issues with lightbox? The close is extremely slow on iOS. It's like the laggiest thing I've ever seen. 😆 

https://github.com/oblador/react-native-lightbox/issues/2
https://github.com/oblador/react-native-lightbox/issues/6

Doesn't seem to be any solid solutions though, unfortunately.

cooperka:
Haha yeah it's really bad for large images. The issue is actually that it's trying to display an enormous image at full resolution instead of down-sampling it.

I wrote custom logic in my client app to resize the image to 720p at 80% quality before displaying, and then it works very smoothly.

booboothefool:
@cooperka Can you explain how you did it?

I'm using https://github.com/ivpusic/react-native-image-crop-picker
```
    CropPicker.openPicker({
      compressImageMaxWidth: 100,
      compressImageMaxHeight: 100,
      compressImageQuality: 0.1,
    })
```
which makes the image look like a butthole, like you can see every single big pixel and can't even tell what the picture is anymore, but I still can't get the animation smooth. 😝

mengpeilee:
Does image can be zoomed now?
I see the PR(zoom image) still open in react-native-lightbox

mafiusu:
Sadly there is not much engagement for react-native-lightbox. Last update was before 7 month.

mengpeilee:
Oh on><
Does there have any way to let zoom image become possible?

PR #305, opened 2016-12-13
Feature/system messages

muszu:
Add ability to display system messages in chat.


README.md
```diff
@@ -87,6 +87,7 @@ See [example/App.js](example/App.js)
     avatar: 'https://facebook.github.io/react/img/logo_og.png',
   },
   image: 'https://facebook.github.io/react/img/logo_og.png',
+  isSystemMessage: false,
```
dgdavid: From my point of view, `fromSystem` would be a better name. `message.fromSystem` seems more readable and less redundant than `message.isSystemMessage`.

src/MessageContainer.js
```diff
@@ -109,13 +109,15 @@ export default class MessageContainer extends React.Component {
       message.user = {};
     }
 
+    let position = message.user._id === this.props.user._id ? 'right' : 'left';
```
dgdavid: It seem to me that this could be improved using an _util_ function. Since if message is from system `user` should be null, you can write a function that receive a message and the current user  such as

```javascript

function messagePosition(message, currentUser) {
  if (message.fromSystem) {
    return 'center';
  }

  return (message.user._id === currentUser._id) ? 'right' : 'left';
}
```

README.md
muszu: You're right. Thank you for your suggestion :)

src/MessageContainer.js
muszu: Fixed :)

src/MessageContainer.js
```diff
@@ -109,13 +117,15 @@ export default class MessageContainer extends React.Component {
       message.user = {};
     }
 
+    const position = messagePosition(message, this.props.user);
```
dgdavid: You don't need a variable here. You can use `messagePosition` directly in  `messageProps`.

src/MessageContainer.js
```diff
@@ -109,13 +117,15 @@ export default class MessageContainer extends React.Component {
       message.user = {};
```
dgdavid: Note that in the `renderRow` method there is a warning when message has not user. If the `message.fromSystem` concept is introduced, you need to review the warning condition.

muszu: I will check all changes over the weekend and provide some screens (and improve style of system message).

muszu:
Sorry for such a long break.
I rebased my branch to be up to date with master branch.
I changed a little bit the look of system messages but I still didn't make a good self-review of my changes.



By the way, could I change double tick to material icon done/done all in another pr?
It will cause an extra dependency, but I think it's worth it.
(https://material.io/icons/#ic_done_all)

src/Bubble.js
```diff
@@ -161,6 +161,21 @@ const styles = {
       borderTopLeftRadius: 3,
     },
   }),
+  center: StyleSheet.create({
+    container: {
+      flex: 1,
```
dgdavid: From `react-native` 0.39, `flex: 1` is not longer necessary. Now, `react-native-gifted-chat` 0.1.0 requieres `react-native` >= 0.40. So, could you please review if this could be removed?

src/Bubble.js
```diff
@@ -60,7 +60,7 @@ export default class Bubble extends React.Component {
     if (this.props.renderTicks) {
         return this.props.renderTicks(currentMessage);
     }
-    if (currentMessage.user._id !== this.props.user._id) {
+    if (currentMessage.fromSystem || (currentMessage.user._id !== this.props.user._id)) {
```
dgdavid: Only as suggestion, maybe is better to extract `(currentMessage.user._id !== this.props.user._id)` to an `utils` function such as `isOwnMessage`. This function could be something like 

```js
function isOwnMessage(props) {
  const { currentMessage, user } = props;
  
  return currentMessage.user._id == user._id;
}
````

And then simply the condition to a

`if (currentMessage.fromSystem || !isOwnMessage(this.props)`

or even create a `isFromOtherUser` method in order to avoid the negation (`!`) inside the conditional.

File: src/Bubble.js [hunk 7]
```diff
@@ -161,6 +161,21 @@ const styles = {
...
+    },
+    wrapper: {
+      borderRadius: 15,
+      backgroundColor: '#7E8B99',
+      marginHorizontal: 30,
+      minHeight: 20,
+      justifyContent: 'flex-end',
+    },
+    containerToNext: {},
+    containerToPrevious: {},
```
dgdavid: Why empty objects? Are you planing to receive their values as a prop?

muszu: At first I thought that messages from system could have 'user' field and handleBubbleToNext/Previous would throw an error.
But after your earlier suggestion that messages from system shouldn't have user field they are unnecessary.
I will fix it and test with new react in ~24h

muszu: Sorry, I just realized how wrong I was. HandleBubbleToNext/Previous would not throw an error even if system messages would have an user.
Unfortunately I didn't manage to test iOS today, but I will be able to do that on monday.

src/MessageContainer.js
```diff
@@ -100,22 +100,32 @@ export default class MessageContainer extends React.Component {
...
+    const position = this.messagePosition(message, this.props.user);
+
     const messageProps = {
       ...this.props,
       key: message._id,
       currentMessage: message,
       previousMessage: message.previousMessage,
       nextMessage: message.nextMessage,
-      position: message.user._id === this.props.user._id ? 'right' : 'left',
+      position,
```
muszu: moved :)

src/Bubble.js
muszu: @dgdavid I checked both platform and `flex: 1` is unnecessary (I have already removed `flex: 1` from Bubble).

src/Bubble.js
muszu: I extracted `isFromOtherUser` to utils.

src/Bubble.js
```diff
@@ -60,8 +60,8 @@ export default class Bubble extends React.Component {
     if (this.props.renderTicks) {
         return this.props.renderTicks(currentMessage);
     }
-    if (currentMessage.user._id !== this.props.user._id) {
-        return;
+    if (isFromOtherUser({ currentMessage, user: this.props.user })) {
```
dgdavid: I see that you're using property/value shorthands for the `currentMessage` taking advantage of ES6's Enhanced Object Literals. IMO, is better to keep coherence. I mean, if you cannot use the shorthands for all properties, is preferable to avoid it since it's a little confusing.

Furthermore, for this kind of functions is more clear to pass it the arguments to compare separately instead of inside an object. However, I think that at this point we should to ping @kfiroo to get his opinion and recommendations.

src/utils.js
```diff
@@ -21,6 +21,12 @@ export function isSameUser(currentMessage = {}, diffMessage = {}) {
 
 }
 
+export function isFromOtherUser({ currentMessage = {}, user = {} }) {
+
+  return currentMessage.fromSystem || (currentMessage.user._id !== user._id);
```
dgdavid: I'm not sure that is a good idea to use `currentMessage.fromSystem` here. Semantically, the function is checking is the given message is from a user different to given one. Even when probably a message from the system will be treated as a message _from other user_ (in fact, it doesn't have user), I believe that is out of the scope of this function to check it.

I prefer to move it off the function and keep it in the original `if` statement, but once again advices from @kfiroo  are welcome. In the end, he is who decides if this feature will be included (or not) and in what way.

kfiroo:
Guys, I'm still not convinced this PR is necessary for the scope of this project. (not trying to step on any toes here)

I have system messages in my app and we didn't need to change any internal code of this library.
I feel like overriding `renderMessage` and rendering whatever you need in there should work just fine for anyone and is flexible enough.

dgdavid:
Thanks @kfiroo .

Then, could you give to @muszu an example? He had been working on this and maybe you can help him to achieve the same or similar result without any modification in the library.

I know that here we're busy, but for this things would be great activate the wiki and put there simple guides for use cases like this.

PR #317, opened 2016-12-21
Add eslint script (extends airbnb + react + react-native)

cooperka:
Currently 486 errors, 7 warnings!

We'll have to go through and iteratively fix all of them. Most popular IDEs have an ESLint plugin where the lints will show up as red underlines, so that makes things much easier. There's also an automatic cleanup with `eslint --fix` that I haven't tried yet.

Some issues should be ignored, and that can be done line-by-line with `// eslint-disable-line rule-name` or for blocks with `/* eslint-disable rule-name */ ...code... /* eslint-enable */`

The README for [eslint-config-cooperka](https://github.com/cooperka/eslint-config-cooperka) has also been updated with config steps in case you want to have a look.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,3 @@
+node_modules
+android
+ios
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,16 +1,12 @@
 {
-  "plugins": [
-    "react",
-    "react-native"
-  ],
-  "extends": ["eslint:recommended", "plugin:react/recommended"],
-  "parserOptions": {
-    "sourceType": "module",
-    "ecmaFeatures" : {
-      experimentalObjectRestSpread: true
-    }
-  },
+  "extends": "cooperka/react-native",
+
   "env": {
-    "node": true
-  }
+    "browser": true,
+    "jest": true
+  },
+
+  // Any rules here will override those from
+  // https://github.com/cooperka/eslint-config-cooperka.
+  "rules": {}
 }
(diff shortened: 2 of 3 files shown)


kfiroo:
@cooperka shouldn't we only have `eslint-config-cooperka` in our devDependencies?

cooperka:
Yeah, good point that would be easier. I'll update it soon.

Should we fix lints in this PR, or make a new one?

kfiroo:
New one i think, it might not be so friendly

cooperka:
@kfiroo apparently npm doesn't work that way... [here's my attempt](https://github.com/cooperka/eslint-config-cooperka/commit/a8f06ad8e25a8a4e7a3ae66fd06ac3e18b991926).

Do you have a different suggestion for how we could use `eslint-config-cooperka` without also specifying the other dependencies such as `eslint-config-airbnb`? It seems like Airbnb has the same problem, hence their [nifty script](https://www.npmjs.com/package/eslint-config-airbnb) for updating automatically (which I included in my own README).

kfiroo:
@cooperka Which version of npm are you using?
This sounds a lot like an npm 2 thing, but I'm not sure.

We can keep the dependencies in devDeps, just feels strange that it might fall out of sync with your code who uses it.

cooperka:
@kfiroo I tried with npm `v4.0.5` on both node `v6.9.1` and `v7.3.0`. Have you tried yourself? I'm guessing there's no easy solution, since Airbnb couldn't even solve it.

kfiroo:
@cooperka sorry I haven't tried it yet :(
But I think you are right, I just thought you forgot to remove them this is why I asked in the first place

So, I'm merging?

cooperka:
Sure! Assuming you're happy with the rules (but we can always change / override them later as we fix lints).

PR #319, opened 2016-12-22
move all exports into index.js

mauron85:
Allows selective imports (tree shaking).

before `import GiftedChat from 'GiftedChat/src/GiftedChat'` imported all react classes.

after: `import GiftedChat from 'GiftedChat/src/GiftedChat'` only imports GiftedChat class.

This allows selectively replace GiftedChat components for alternative implementations, without need to fork whole project.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,1 +1,32 @@
-export * from './src/GiftedChat';
+import GiftedChat from './src/GiftedChat';
+import Actions from './src/Actions';
+import Avatar from './src/Avatar';
+import Bubble from './src/Bubble';
+import MessageImage from './src/MessageImage';
+import MessageText from './src/MessageText';
+import Composer from './src/Composer';
+import Day from './src/Day';
+import InputToolbar from './src/InputToolbar';
+import LoadEarlier from './src/LoadEarlier';
+import Message from './src/Message';
+import MessageContainer from './src/MessageContainer';
+import Send from './src/Send';
+import Time from './src/Time';
+import * as utils from './src/utils';
+
+export {
+  GiftedChat,
+  Actions,
+  Avatar,
+  Bubble,
+  MessageImage,
+  MessageText,
+  Composer,
+  Day,
+  InputToolbar,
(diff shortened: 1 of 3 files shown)


kfiroo:
@mauron85 will this break the code for current users?

mauron85:
Absolutely not. You have my word :-)

mauron85:
To give some more insights why is this important for me.

Imagine you want to replace Bubble class with your own implementation

```
export default class Query extends Component {
...
}
```

And use this extended class in your app

```
import GiftedChat from 'react-native-gifted-chat/src/GiftedChat';

export default class Chat extends Component {
...
  renderBubble(props) {
    switch (props.currentMessage.type) {
      case 'query':
        return (
          <Query {...props} onResponse={this.onResponse} />
        );
      default:
        return (
          <Message {...props} onPress={this.onMessagePress} />
        );
    }
  }

  render() {
    const { user, messages } = this.props;

    return (
      <View style={styles.container}>
        {(() => {
          if (!this.props.connected) {
            return <ErrorBar>Not connected</ErrorBar>;
          }
        })()}
        <GiftedChat
          user={user}
          messages={messages}
          onSend={this.onSend}
          renderBubble={this.renderBubble}
        />
      </View>
    );
  }
}
```

The issue of current implementation (small and project is awesome) is that Bubble class was imported, but not used.

kfiroo:
Thanks! that's much clearer.
Unfortunately, your change will not solve the problem you are trying to solve :(
Think about it, `Bubble` is still in the require chain.
`GiftedChat` -> `MessageContainer` -> `Message` -> `Bubble`
You only changed the way it was exported

mauron85:
It was just an example. In reality I'm only extending Bubble (not replacing it). There are no downsides of refactoring exports into index.js file, so I still recommend it.

kfiroo:
@mauron85 Sure let's do it.
Just wondering if there are any upsides to it as well :)

PR #322, opened 2016-12-29
Removes unnecessary flex: 1 in bubble styles

kentos:
Since react 0.39 fixes some "broken" flexbox behaviour these arent needed anymore.

### DIFF
--- a/src/Bubble.js
+++ b/src/Bubble.js
@@ -122,7 +122,6 @@ export default class Bubble extends React.Component {
 const styles = {
   left: StyleSheet.create({
     container: {
-      flex: 1,
       alignItems: 'flex-start',
     },
     wrapper: {
@@ -141,7 +140,6 @@ const styles = {
   }),
   right: StyleSheet.create({
     container: {
-      flex: 1,
       alignItems: 'flex-end',
     },
     wrapper: {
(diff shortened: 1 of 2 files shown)


dgdavid:
As I said in https://github.com/FaridSafi/react-native-gifted-chat/pull/327, IMO is better to review the breaking changes and fix them in a specific PR. But maybe this means to force everybody to update their apps to the last version of react-native in order to keep using RNGC or at least its more recent version.

@kfiroo, How manage this?

kfiroo:
@dgdavid That's actually a very good question, I have no idea :)
I think we should try to keep this repo compatible with the latest react-native version.

If someone uses an earlier version of RN he could use an earlier version of RNGC

What do you think?

tafelito:
I tried to upgrade to 0.39 and it was breaking in Android. They fixed the issue in 0.39.2

Ideally, I think its better to support from a version and up, so for people who still on older versions for different reasons, they can still use it. But that's not always the case.

If you try to run with version 0.40 then it beaks, so it's hard to keep up to date with every latest release, unless you can fix all the issues as soon as a new RN release comes up.

kfiroo:
So maybe 1 version below the latest?

kfiroo:
@kentos I cannot accept this PR with the new commit :(
The `InteractionManager` is still needed there to prevent slow transitions to the chat, afaik

kentos:
@kfiroo oooh snap. wasnt meant to be in the PR :/ I messed it up. I'll remove the PR as we're using our own forked for the moment being.

PR #327, opened 2017-01-04
Remove touch events and external keyboard dependency

tafelito:
I removed the dependency for `react-native-dismiss-keyboard` and removed the touch events. Instead of using touch events and dismiss manually the keyboard, we can remove `keyboardShouldPersistTaps` and it has the same effect.

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -7,8 +7,8 @@
     "sync": "rm -rf ./node_modules/react-native-gifted-chat; sane '/usr/bin/rsync -v -a --exclude .git --exclude example --exclude __tests__ --exclude node_modules ../ ./node_modules/react-native-gifted-chat/' .. --glob='{**/*.json,**/*.js}'"
   },
   "dependencies": {
-    "react": "15.4.1",
-    "react-native": "0.39.0",
+    "react": "15.3.2",
+    "react-native": "0.37.0",
     "react-native-camera-roll-picker": "^1.1.7",
     "react-native-gifted-chat": "file:../",
     "react-native-nav": "^1.1.4"
--- a/example/package.json
+++ b/example/package.json
@@ -7,8 +7,8 @@
     "sync": "rm -rf ./node_modules/react-native-gifted-chat; sane '/usr/bin/rsync -v -a --exclude .git --exclude example --exclude __tests__ --exclude node_modules ../ ./node_modules/react-native-gifted-chat/' .. --glob='{**/*.json,**/*.js}'"
   },
   "dependencies": {
-    "react": "15.2.1",
-    "react-native": "^0.30.0",
+    "react": "15.4.1",
+    "react-native": "0.39.0",
     "react-native-camera-roll-picker": "^1.1.7",
     "react-native-gifted-chat": "file:../",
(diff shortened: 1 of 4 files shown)


tafelito:
I also changed RN version, but there was an error 0.39 in Android so I downgraded to 0.37 again. I already tested in 0.39.2 and it's working for both iOS and Android.

kfiroo:
@dgdavid You are right, however it's only the example deps.
@tafelito If you are down to split the PRs that would be awesome, otherwise I think it's fine

tafelito:
@dgdavid @kfiroo The reason why I upgraded the example is because it was not working for me on the previuos version. I read of the issues posted here that upgrading to 0.37 and 15.3.2 solved the issue so I did that. 

With 0.30 there is this error

```
...react-native-gifted-chat/example/node_modules/react-native/Libraries/WebSocket/RCTSRWebSocket.m:1334:5: error: ignoring return value of function declared with warn_unused_result attribute [-Werror,-Wunused-result]
    SecRandomCopyBytes(kSecRandomDefault, sizeof(uint32_t), (uint8_t *)mask_key);
    ^~~~~~~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2 errors generated.
```

Any ideas how to solve that using 0.30?

tafelito:
@kfiroo I saw you merged a PR where you use the keyboardShouldPersistTaps={always} and I wanted to ask you the reason behind this. We are still using the dependency on a third party plugin to dismiss the Keyboard while there is no need if you don't set that property at all. Maybe I'm missing something but I cannot see what. 

Thanks

kfiroo:
@tafelito Hey! 
Yes, sorry, I didn't realise the connection between the PRs, honestly I like your fix better.

I see that in your PR that you removed the `keyboardShouldPersistTaps` prop completely, why?

tafelito:
@kfiroo if you don't set the keyboardShouldPersistTaps, it means that it will hide the keyboard when you touch the scrollview component. 

As it says it the docs 
`'always', the keyboard will not dismiss automatically, and the scroll view will not catch taps, but children of the scroll view can catch taps`

So it means that the Keyboard will not be dismissed if you tap the list. INHO, I don't think its a good idea. 

So pretty much, If you don't set the value (or you set it as 'never'), it will hide the keyboard when you tap the list, but not when you send the a message.

kfiroo:
@tafelito Whatsapp has 2 different behaviours for android and ios.
Android looks like its is set to 'never' - you can scroll the list and tap on it and the keyboard stays.
iOS will leave the keyboard open when scrolling but closes it when you tap the list.

I think it's bet to leave this decision to the users of this package
What do you think?

tafelito:
@kfiroo So I guess in that case yes, I agree with you, we can leave a default but allow the user to set a different value if they want. To me the iOS behavior looks more appropriate, but as long as the user can change the default value, I think either of them it should be ok

tafelito:
@kfiroo I'm closing this PR and creating a new one based on what we discussed earlier. And without any RN version change. At least not in this PR.

PR #329, opened 2017-01-05
Remove runAfterInteractions block on layout calculation

tobycox:
We were seeing the block in `runAfterInteractions` in `onLayout` never being called.
This would mean none of the chat UI would render.

Replicated this issue in react-native 0.36 and 0.38
It was only sporadic and seemed to happen more often on device.

I don't know too much about `runAfterInteractions`, is it guaranteed to be run?

Removing this block fixed our issue and didn't seem to break anything (our navigator animation is still smooth)

### DIFF
--- a/src/GiftedChat.js
+++ b/src/GiftedChat.js
@@ -439,13 +439,11 @@ class GiftedChat extends React.Component {
         onLayout={(e) => {
           const layout = e.nativeEvent.layout;
           this.setMaxHeight(layout.height);
-          InteractionManager.runAfterInteractions(() => {
-            this.setState({
-              isInitialized: true,
-              text: '',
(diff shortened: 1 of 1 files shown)


kfiroo:
This line is meant to hold off any rendering until the current "interactions" are over, so if you are navigating to a page with `GiftedChat` the transition would be smooth.
Not sure it's safe to remove

tobycox:
@kfiroo Yep, you're right. This isn't the right approach. I dug a bit (a lot) deeper and found that all `runAfterInteractions` were being blocked because we weren't clearing up the looping animation from the loading screen in our app.

So for anyone encountering the same kind of problem: make sure you clean up your animations properly.

PR #332, opened 2017-01-09
Bug fix/update keyboard property

charlle:
deprecated property true for keyboardShouldPersistTaps

### DIFF
--- a/src/GiftedChat.js
+++ b/src/GiftedChat.js
@@ -68,7 +68,7 @@ class GiftedChat extends React.Component {
 
     this.invertibleScrollViewProps = {
       inverted: true,
-      keyboardShouldPersistTaps: true,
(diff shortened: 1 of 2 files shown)


kfiroo:
@charlle Hey! Thanks for the PR!
Quick question, will this break earlier versions of RN?

charlle:
Yes.  Here is more info: https://github.com/facebook/react-native/pull/10628.  True will still be a valid value, but will throw a warning for now.  Here is the change to the core component ScrollView: https://github.com/facebook/react-native/commit/d56530d7d5f20f84c45a7187ce564c03a023c6ae.
previously expecting boolean, so will break older versions of RN when receive string.

kfiroo:
@charlle Cool, merging

PR #335, opened 2017-01-18
custom ListView props support

luhui:
for some purpose, we want to listen event on ListView, like `onScroll`, `onLayout`.... 
such as animate a view when list view in scroll


kfiroo:
@luhui Hey, thanks for the PR.
Please update the readme to reflect your change.
Also, I'm a bit concerned users might override props used by `gifted-chat` and break functionality what do you think?
I would also add `defaultProps` for it as an empty object

luhui:
@kfiroo Hi, thanks for your advice
code and README has update

PR #340, opened 2017-01-18
Update RN version to 0.40

tafelito:
@kfiroo So in order to get the new keyboard props working, we need to update to the latest RN version. I managed to build GC with the latest RN making some changes on the xcode project. 

Here are all the changes we need to make it work with RN 0.40

I'm not sure how you want to handle this, I just create the PR, let me know if you think we should do something else.

This includes the changes in the info.plist to make the camera work as well

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -7,8 +7,8 @@
     "sync": "rm -rf ./node_modules/react-native-gifted-chat; sane '/usr/bin/rsync -v -a --exclude .git --exclude example --exclude __tests__ --exclude node_modules ../ ./node_modules/react-native-gifted-chat/' .. --glob='{**/*.json,**/*.js}'"
   },
   "dependencies": {
-    "react": "15.2.1",
-    "react-native": "^0.30.0",
+    "react": "~15.4.0-rc.4",
+    "react-native": "^0.40.0",
     "react-native-camera-roll-picker": "^1.1.7",
     "react-native-gifted-chat": "file:../",
     "react-native-nav": "^1.1.4"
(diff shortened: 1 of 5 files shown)


tafelito:
we could create a new branch if you want, so we don't have to merge straight into master, and then it'll be easier to manage different release versions for GC

example/package.json
```diff
@@ -7,8 +7,8 @@
     "sync": "rm -rf ./node_modules/react-native-gifted-chat; sane '/usr/bin/rsync -v -a --exclude .git --exclude example --exclude __tests__ --exclude node_modules ../ ./node_modules/react-native-gifted-chat/' .. --glob='{**/*.json,**/*.js}'"
   },
   "dependencies": {
-    "react": "15.2.1",
-    "react-native": "^0.30.0",
+    "react": "~15.4.0-rc.4",
```
kfiroo: Is this necessary?

tafelito: RN has the dependency on that specific react version. I tried to use the latest React version 15.4.2 but it does not work. I think when they release the latest RN to npm they'll update that too. I hope so

tafelito: you can see the peerDependency here [https://github.com/facebook/react-native/blob/master/package.json](https://github.com/facebook/react-native/blob/master/package.json)

sibelius: https://github.com/ncuillery/rn-diff can help as well

kfiroo:
@tafelito This is only for the example project right?

tafelito:
yes @kfiroo, only for the example project. GC does not have any hard dependency on RN

Issue #350, opened 2017-01-27
Drawer and deep linking


razvanpaul:
I use this amazing library in coalition with [react-native-navigation](https://github.com/wix/react-native-navigation). It seems to work fine in some cases, even if I had to implement some handlers. See #270 and #244.

I have the following: Homepage, Drawer, Match and finally the Chat. At the initial moment of the app, I present the homepage with the drawer in the left side, where I choose the Match page. The Match is instantiated using deep links, `onNavigatorEvent` and `this.props.navigator.resetTo`.

Finally, when an item is tapped in the Match page, I want the app to go to the Chat page, but all I get is this:
![simulator screen shot 27 jan 2017 15 01 27](https://cloud.githubusercontent.com/assets/8782666/22371741/7fac34cc-e4a1-11e6-9c80-003d13a3fde2.png)

I use `this.props.navigator.push` to do it, but I've also tried `this.props.navigator.showModal`, nothing worked.
Observation: Everything is good when I firstly instantiate the Match page, instead of the Homepage. Apparently something is not ok with react-native-navigation and deep linking here.

Can you tell me any solution please?

* React Native version: 0.39
* react-native-gifted-chat version: 0.0.10
* Platform(s) (iOS, Android, or both?): iOS

razvanpaul:
Update

Apparently, this is the difference when the Match page is firstly instantiated and when it is not:

Instantiated:
![simulator screen shot 27 jan 2017 18 58 52](https://cloud.githubusercontent.com/assets/8782666/22379814/9b7644fa-e4c3-11e6-9e58-25e0f66b89b8.png)


Not instantiated:
![simulator screen shot 27 jan 2017 19 02 49](https://cloud.githubusercontent.com/assets/8782666/22379825/a651c82c-e4c3-11e6-8b8a-bd18fac2ee7c.png)

kfiroo:
Hey @razvanpaul 
Can you try to use latest commit from master and see if your issues are solved?

razvanpaul:
I've tried to get 0.0.11, but it says it's unavailable. How can I get it?

kfiroo:
@razvanpaul in your package.json use this:

```
"react-native-gifted-chat": "git+https://github.com/FaridSafi/react-native-gifted-chat#d3387e2f55a6067beec08d21eed4f42402e7779e",
```

razvanpaul:
@kfiroo Thank you!

[...]

PR #355, opened 2017-01-30
InteractionManager breaks sometimes, causes hard-to-debug issues

razvanpaul:
This pull request solves #350. Basically, when I was using react-native-navigation and calling a function called `resetTo` there, it simply broke the `InteractionManager` functionality, it was simply not calling the function you were passing as a parameter. More info about that [here](https://github.com/facebook/react-native/issues/8624).

Speaking about GiftedChat, I believe it's a serious problem and we should basically always display the chat. In my case, I had to debug the core of two big React Native libraries, so it would be good to merge this pull request.

A video about my bug: http://sendvid.com/qtsi7t1b
Solved now. Hope it helps you too.

### DIFF
--- a/src/GiftedChat.js
+++ b/src/GiftedChat.js
@@ -26,6 +26,7 @@ import MessageContainer from './MessageContainer';
 import Send from './Send';
 import Time from './Time';
 import GiftedAvatar from './GiftedAvatar';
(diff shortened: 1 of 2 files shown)


Braidio:
This should definitely be added to the main repo. 

+1

PR #369, opened 2017-02-10
Added "renderAvatarOnTop" prop to Avatar.js

gperdomor:
Currently, the Avatar component render the avatar of a message on the last consecutive message from the user.

This prop allows users to change this behavior, so that the avatar is rendered on the first message.

The default value is false, which keeps the current behavior.


kfiroo:
@gperdomor Looks good, thanks

Issue #377, opened 2017-02-22
Blank screen during transition


iclems:
#### Issue Description

Also see: https://github.com/aksonov/react-native-router-flux/issues/1647

During a transition from a list view into a gifted chat view, the view is not rendered until after the component has mounted and is in center stage. As a result, during the transition the screen remains blank. This is not related to the state being set in the constructor or componentWillMount. The input is not even visible.

#### Steps to Reproduce / Code Snippets

react-native-router-flux scenes where the destination is a GiftedChat

#### Expected Results

The GiftedChat should ideally render before the transition to be visible. At least parts of the UI should be there. 

#### Additional Information

* react-native-router-flux v3.37.0
* react-native v0.41.2
* react-native-gifted-chat version: 0.1.0
* Platform(s) (iOS, Android, or both?): both

kfiroo:
@iclems This is a know issue with `gifted-chat`, it's been here for a while now, I hope we can get to it soon.
PR is welcome, btw :)

As you can see [here](https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/GiftedChat.js#L438) the first render is always an empty view that waits for the transition to finish before measuring and rendering the chat content.

I believe it was meant to prevent frame drops when animating a lot of content

iclems:
Thank you for pointing me in the right direction. I will test and figure out what's that helping with. I guess using a debounce you could render once before and then wait until it's complete to render again.

iclems:
Interesting. I just ran tests removing the InteractionManager.runAfterInteractions wrapper and executing the setState directly and it works perfectly now. I assume this would only be required if you do add items while the animations is ongoing. I will test this in particular an report. If so, the trick could be to render once before the animation and delay any further state update until after.

iclems:
To give context on alternative solutions, what other (non react native apps) do when the next view is resource intensive is pause for 100ms before starting the animation. I just can't figure out yet, especially using router-flux how you could "prepare" the component before starting the transition.

kfiroo:
@iclems Thanks for the research! :)

What react-native has to offer in this area is pretty cool, the `ListView` itself, that holds the messages, has `initialListSize` and `pageSize` props which tell it how many items it should render on the first render and how many on subsequent renders.

If everything works fine, it should render `initialListSize` items when it mounts and the transition should go smoothly.

[...]

Issue #228, opened 2016-09-08
Transitions on iOS are extremely glitchy when moving to gifted-chat


ghost:
#### Issue Description

When navigating the a page containing the `<GiftedChat` component the transition is extremely buggy, the page behind moves over slightly to the left and then after 2-3 seconds the chat component instantly flashes onto the screen. This is happening with every animation type under SceneConfigs in which there is a movement transition. I have not tested with `FadeAndroid`. 

The Footer for the `<GiftedChat` also does not occur until this delay is finished.

NOTE: I added a custom component on the top of the screen with a fixed height and absolute positioning of 64 (A navigation bar) and it transitions over smoothly, it's only the `<GiftedChat` component with a problem.

NOTE2: A blank array of messages on start-up time does not fix the issue, it takes roughly 600ms after the transition completes to snap into place. 
#### Steps to Reproduce / Code Snippets

Just use the example code with React-natives `Navigator` component. Move to it using `push({...})` and the problem occurs.
#### Expected Results

A smooth transition
#### Additional Information
- React Native version: 0.32.0
- react-native-gifted-chat version: 0.0.10
- Platform(s) (iOS, Android, or both?): iOS (Simulator and Device)

FaridSafi:
Hello,
I suggest to try https://facebook.github.io/react-native/docs/performance.html#slow-navigator-transitions in your container
Also, you can consider testing on a device with Release build to increase performance

jonathanewerner:
I experience the same issue. I have implemented the linked performance tweets and the problem appears in a release build. 
It also appears when showing the chat page initially (without transitioning to it): it takes about 600ms to render the chat bubbles, resulting in a very jerky experience. 
Any pointers how to tackle this? I'd be happy to PR but a direction which component might introduce this delay would be helpful.

stantoncbradley:
@jonathanewerner can you elaborate on how you solved this? Where did you move the state the was set inside `runAfterInteraction`?

kfiroo:
Maybe related https://github.com/FaridSafi/react-native-gifted-chat/issues/377

cooperka:
This is hopefully fixed now on master. As @iclems [mentioned here](https://github.com/FaridSafi/react-native-gifted-chat/issues/377#issuecomment-281961674), the change is a good workaround that should make things better for most people, but it could lead to slower animations in some cases. PRs to improve performance are always welcome!

Please try it out and let us know if the new code didn't actually resolve this particular issue.

[...]

PR #380, opened 2017-02-23
Removed InteractionManager.runAfterInteractions during init

iclems:
Solves https://github.com/FaridSafi/react-native-gifted-chat/issues/377

### DIFF
--- a/src/GiftedChat.js
+++ b/src/GiftedChat.js
@@ -1,7 +1,6 @@
 import React from 'react';
 import {
   Animated,
-  InteractionManager,
   Platform,
   StyleSheet,
   View,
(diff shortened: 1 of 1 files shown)


stantoncbradley:
also solves #228.  would love to see this merged so we can use upstream branch again

boennemann:
We're using this commit cherry-picked onto 0.1.3 and it's working really well.
It solved a bug where GiftedChat didn't show up at all in some cases (uninitialized).
In the cases where it did show up it took a really long time to load, and it turns out that now even these cases are fast.

Also we've seen problems with `runAfterInteractions` on Android before, so this will likely benefit Android as well.

We would love to see this land on master 👍

iclems:
I haven't fully tested that yet but I suspect with the new React Navigation project being fully native for transitions, there shouldn't be such issues anymore.

kfiroo:
@iclems Hey, sorry for the delay, I was away for a month.
Can you please fix the conflicts so I can merge this?
Thanks! :)

cooperka:
Ping @iclems this would be great to have!

bporter2387:
Idk if I'm doing something wrong, but I can't get the Chat to render unless I remove the following from onInitialLayoutViewLayout() 
`if (layout.height <= 0) {
      return;
    }`
The layout always comes out 0. I am using conditional component rendering using a switch statement, so maybe that's why? So if 'activeTab == "chat"' then render the giftedChat. I'd more than happy to create a PR with the change if removing that is okay with all of you.

cooperka:
@bporter2387 if you're still having the problem you mentioned above, would you mind opening an issue or a PR so we can discuss in more detail?

PR #382, opened 2017-02-24
Reset Input Toolbar can over calculate from bottomOffset props

erenn16:
Ran into an issue in my implementation where I have a bottom offset for the input tool bar in the chat.  When a user inputs a message, dismisses the keyboard and then sends the message, the text input tool bar would end up being rendered too low and behind a few other view components.  Instead of using the this.props.bottomOffset, it should be using this.getBottomOffset() so that it doesn't incorrectly position when the keyboard is dismissed.  Verified that sending a message with the keyboard showing still works as intended.


kfiroo:
@erenn16 Thanks

PR #387, opened 2017-03-03
Allow custom styling of buttom container of bubble

brendan-rius:
In the bubble, there is a container that holds both the time, and the ticks. This container did not provide customization options and forced both the ticks and the time to be rendered on the right (with `flex-end`).

I created another prop, `bottomContainerStyle` that allows to customize the style of the container :)


kfiroo:
Thanks

PR #392, opened 2017-03-07
Export MessageContainer in GiftedChat

brendan-rius:
It was not exported making it harder to customize GifterChat

### DIFF
--- a/src/GiftedChat.js
+++ b/src/GiftedChat.js
@@ -537,6 +537,7 @@ export {
   InputToolbar,
   LoadEarlier,
   Message,
+  MessageContainer,
   Send,
   Time,
   GiftedAvatar,


kfiroo:
@brendan-rius Hey, I'm wondering, how would you use this?

brendan-rius:
I needed to extend the GiftedChat class directly. Thus I copied the original code of GiftedChat but I needed to import `MessageContainer` to do this. It was not exported, making it impossible to do something like:

```javascript
import {
    Actions,
    Avatar,
    Bubble,
    MessageImage,
    MessageText,
    Composer,
    Day,
    InputToolbar,
    LoadEarlier,
    Message,
    MessageContainer, // <---- here
    Send,
    Time,
    GiftedAvatar,
    utils
} from  'react-native-gifted-chat'

class GiftedChat extends React.Component {
   // THIS CODE USES MessageContainer
}
```

kfiroo:
@brendan-rius Oh, well this is not the common use case :)

Issue #289, opened 2016-11-23
Feature Request: Disable avatars?


f0rr0:
I don't need avatars for my use case. The most immediate way to do so seemed returning `null` from `renderAvatar`. However, this messes up the styles. `avatars` or `showAvatars` prop might be handy.
![simulator screen shot nov 23 2016 4 11 47 pm](https://cloud.githubusercontent.com/assets/8574219/20558743/8ec211e4-b197-11e6-9ecb-18475439d8c9.png)

f0rr0:
On another note, why is `Avatar` rendered if I had `renderAvatar={() => null}` on the component

dgdavid:
This is happening as a consequence that [this piece of code](https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/Avatar.js#L25-L31) used to avoid render the avatar in consecutive messages of an user in the same day (but keeping the gap/alignment).

In fact, as we can see in your attached picture, the last message (which doesn't have `nextMessage`) is using given `renderAvatar` function.

I think that needs a refactorization. If I find some free time, and if nobody make it before, I'll try to fix it and make a PR.

dgdavid:
Mmm, I'm not sure right now, but reviewing https://github.com/FaridSafi/react-native-gifted-chat/pull/234 it seems that could solve this.

grin:
The option to disable avatars and initials would be really helpful...I also don't need it in my app.

grin:
Implemented in https://github.com/FaridSafi/react-native-gifted-chat/pull/404

kaberibasu:
Not able to resolve the problem.
`renderAvatar={() => null}` does not work.

Faolain:
As noted in the documentation

```renderAvatar (Function) - Custom message avatar; set to null to not render any avatar for the message```

@f0rr0 can you please close this issue?

PR #404, opened 2017-03-21
Disable avatars if user provided renderAvatar function returns null

grin:
Implements https://github.com/FaridSafi/react-native-gifted-chat/issues/289


grin:
Ended up not needing this...

grin:
I left the fix on a separate branch (if anyone is interested): https://github.com/grin/react-native-gifted-chat/commits/bg/disable_avatars

Issue #407, opened 2017-03-26
Undefined is not an object exception compiling the example project


jeromegrosse:
#### Issue Description

Exception thrown with the vanilla example of the project:
```
TaskQueue: Error with task : undefined is not an object (evaluating 'this.props.bottomContainerStyle[this.props.position]')
```
This evaluation is within the [Bubble.js file](https://github.com/FaridSafi/react-native-gifted-chat/blob/2914239544decbc3d208848ade8666833ffa12fa/src/Bubble.js#L132).

This appears to have been introduced by  [this commit](https://github.com/FaridSafi/react-native-gifted-chat/commit/9d56c4a33b3a43c9e4f0481eaaf483f759d3474f)

#### Steps to Reproduce / Code Snippets

Clone the repository, compile the example

#### Expected Results

Crash of the application returning the aforementioned error.

#### Additional Information

* React Native version: 0.40.0
* react-native-gifted-chat version: 0.1.3
* Platform(s) (iOS, Android, or both?): Android (both, I guess?)

kfiroo:
@brendan-rius can you take a look please? 
I'd hate to revert that commit..

PR #408, opened 2017-03-26
Bubble.js: Default value for prop bottomContainerStyle

jeromegrosse:
This solves the following issue: #407 

The reason of the bug was that there wasn't any default value for the newly introduced `bottomContainerStyle` property and therefore `this.props.bottomContainerStyle` is undefined if nothing is passed to the component, generating the mentioned exception when evaluating `this.props.bottomContainerStyle[this.props.position]`.


kfiroo:
@jeromegrosse Thanks :)

Issue #413, opened 2017-03-30
Long message > Rotate screen > Send button gone offscreen


cooperka:
#### Issue Description

Send button goes off-screen after typing long message then dismissing keyboard [Android only] or rotating to change orientation [Android and iOS] as of latest master.

Send button would normally be on the right in this screenshot, but now it's pushed down below the screen because the text is actually more than just 2 lines:

![screen shot 2017-03-30 at 11 52 07](https://cloud.githubusercontent.com/assets/2047062/24513633/b048223e-153f-11e7-9f78-517d6096236e.png)

#### Steps to Reproduce

(You can use the example app)

1. Type a long sentence with 3+ lines of text
2. Everything should be fine, it works great
3. Dismiss the keyboard or rotate to change device orientation
4. Now you can't see the wrapped text and the send button is off-screen

#### Additional Information

* React Native version: 0.42
* react-native-gifted-chat version: latest master [[April 12](https://github.com/FaridSafi/react-native-gifted-chat/commits/master@%7B2017-04-12%7D)]
* Platform(s): Android

I'm happy to make a PR eventually but that might not happen for a few weeks.

kfiroo:
@cooperka Any idea what's going on there?

cooperka:
@kfiroo I think the Send button is just centering vertically within the very tall view, but the view is getting clipped because some state is lost ("current number of lines" is probably stored somewhere and gets reset on orientation change). If you have around 2-3 lines of text, you can see the button peeking up at the very bottom of the screen. It's centering properly, the view just doesn't have enough room to show everything.

kfiroo:
@cooperka Haven't tested it, but I think I found our criminal: https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/InputToolbar.js#L73
If we change it to `flex-start` if should solve this issue.

cooperka:
(Fixed by #425)

PR #425, opened 2017-04-14
Fix orientation change bugs

cooperka:
This combines the fix from PR https://github.com/FaridSafi/react-native-gifted-chat/pull/280 (rebased because of merge conflicts, and @psrustik has been unresponsive) with a fix for issue https://github.com/FaridSafi/react-native-gifted-chat/issues/413.

Brief summary of issues fixed:

1. On current master, iOS screen turns blank after an orientation change, and Android jumps all over the place if the keyboard is shown before rotation
2. After the fix for orientation changes, composer height gets reset and send button goes off-screen during orientation changes (both iOS and Android)
3. Lots of duplicated math, I cleaned it up and put it into helper methods

Please see individual commits to get a better sense of the changes. Each commit is very simple on its own, though the full PR diff might be confusing to view all at once.

Tested thoroughly on iOS and Android; please also verify yourself if you can :) Basically just type some long text and rotate the phone / dismiss the keyboard / etc. to try to break it.

### DIFF
--- a/src/GiftedChat.js
+++ b/src/GiftedChat.js
@@ -209,9 +209,18 @@ class GiftedChat extends React.Component {
     return composerHeight + (this.getMinInputToolbarHeight() - MIN_COMPOSER_HEIGHT);
   }
 
-  /** Returns the height, based on current window size, without taking the keyboard into account. */
-  getBasicMessagesContainerHeight() {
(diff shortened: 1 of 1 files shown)


kfiroo:
@cooperka Sweet! Thanks! :)

PR #433, opened 2017-04-17
Add maxLength parameter for Composer input

andypmw:
Hi there!
I am working a React Native app that have chat feature.
This app have maximal chat text length specification (in my case is 400 chars).

So I have modify so user can set maxLength property to Composer's TextInput.


kfiroo:
@andypmw Thanks :)

Issue #445, opened 2017-05-01
If renderAvatar is set to not render anything, then the previous messages from the same user are not aligned properly


varungupta85:
#### Issue Description

If a user wants to not show the avatar for the users in the conversation, there is a prop `renderAvatar` which can be set to not return any component which would not render any avatar for the user. But the previous messages from the same user are still shifted towards the right to align the previous messages with the last one if there were an avatar. See the screenshot below.

![screen shot 2017-05-01 at 11 09 43 am](https://cloud.githubusercontent.com/assets/7475133/25573052/3685dfd4-2e5f-11e7-90ed-1c1038a81820.png)

The previous messages from the user should be aligned all the way to the left if `renderAvatar` is not rendering anything.

#### Steps to Reproduce / Code Snippets

Don't render avatars for the the users by setting `renderAvatar = {_ => null}` in the `GiftedChat` component. Send more than one message from a participant and observe the issue.

#### Expected Results

All the messages should be aligned properly

#### Additional Information

* React Native version: 0.42.0
* react-native-gifted-chat version: 0.1.4
* Platform(s) (iOS, Android, or both?): Both

PR #447, opened 2017-05-01
fix-extra spacing shown when avatar is not rendered

varungupta85:
Related to issue #445 

I have added a new prop `shouldRenderAvatar` to control whether a user wants to render an avatar or not. I will document this property if the maintainers agree with the solution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ See [example/App.js](example/App.js)
 - **`isLoadingEarlier`** _(Bool)_ - display an ActivityIndicator when loading earlier messages
 - **`renderLoading`** _(Function)_ - render a loading view when initializing
 - **`renderLoadEarlier`** _(Function)_ - render the load earlier button
-- **`renderAvatar`** _(Function)_ - renders the message avatar
+- **`renderAvatar`** _(Function)_ - renders the message avatar. Set to `null` to not render any avatar for the message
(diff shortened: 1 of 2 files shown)


varungupta85:
I am okay with that. I will make the change.

varungupta85:
@cooperka I have made the change you requested and also updated the README.

cooperka:
I haven't tested on a device, but code LGTM. I also don't have the ability to merge, you'll have to wait for a collaborator for that.

varungupta85:
@cooperka Thanks for reviewing. I have tested it on both Android and iOS devices on my side. I will wait for a collaborator to merge the changes.

cc @kfiroo @FaridSafi

thebigredgeek:
When will this be released?

henrythach:
This breaks the default behavior of showing avatars if developers don't specify a function for `renderAvatar`.  This is because `renderAvatar` is set to `null`, by default, in GiftedChat.js ([See here](https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/GiftedChat.js#L496)).

cooperka:
@henrythach I have the same problem; simple fix is defaulting `renderAvatar` to `undefined` instead of `null`. It's strange that these are all manually defaulted to `null`, I'm not sure what benefit that would give. I'll open a PR to fix this bug.

yurykorzun:
When will this fix be published to the npm registry?

PR #448, opened 2017-05-02
Allow user name to be shown for messages on the left side

varungupta85:
Sometimes just the avatar is not enough to figure out who sent the messages on the left side. Also, the user may not want to render the user avatar on the left also in which case it is even more important to be able to show the person name who sent the messages. Added two new props `shouldRenderUsername` and `renderUsername` to control showing user names for messages.

### DIFF
--- a/src/GiftedAvatar.js
+++ b/src/GiftedAvatar.js
@@ -14,17 +14,7 @@ import {
 // handle only alpha numeric chars
 
 export default class GiftedAvatar extends React.Component {
-  setAvatarColor() {
-    const userName = this.props.user.name || '';
-    const name = userName.toUpperCase().split(' ');
-    if (name.length === 1) {
-      this.avatarName = `${name[0].charAt(0)}`;
-    } else if (name.length > 1) {
-      this.avatarName = `${name[0].charAt(0)}${name[1].charAt(0)}`;
-    } else {
-      this.avatarName = '';
-    }
-
+  static getColor(userName) {
     let sumChars = 0;
     for(let i = 0; i < userName.length; i++) {
       sumChars += userName.charCodeAt(i);
@@ -42,7 +32,18 @@ export default class GiftedAvatar extends React.Component {
       '#2c3e50', // midnight blue
     ];
 
-    this.avatarColor = colors[sumChars % colors.length];
+    return colors[sumChars % colors.length];
+  }
+
+  static getInitials(userName) {
+    const name = userName.toUpperCase().split(' ').filter(word => word !== '');
+    if (name.length === 1) {
+      return `${name[0].charAt(0)}`;
+    } else if (name.length > 1) {
+      return `${name[0].charAt(0)}${name[1].charAt(0)}`;
+    } else {
+      return '';
+    }
   }
 
   renderAvatar() {
@@ -69,7 +70,7 @@ export default class GiftedAvatar extends React.Component {
   renderInitials() {
     return (
       <Text style={[defaultStyles.textStyle, this.props.textStyle]}>
-        {this.avatarName}
+        {GiftedAvatar.getInitials(this.props.user.name)}
       </Text>
     );
   }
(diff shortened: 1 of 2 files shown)


varungupta85:
Below is a screen capture of how the user name appears

screen shot 2017-05-02 at 9 35 29 am

varungupta85:
I have opened a some PRs in the last few days. Could someone please review them at their convenience? Thanks!

cc @kfiroo @cooperka

cooperka:
Hi @varungupta85, thanks for the PR!

Personally, I think if you want to customize the rendering like this you should just pass your own method to `renderBubble`. There's also an `onPressAvatar` method which I personally am using in my app to show a modal popup with more information about the user.

Customization is great, but we also need to think about long-term maintainability, and some features like this may not be worth the added maintenance cost. I'd love to hear other people's thoughts on this.

varungupta85:
Hi @cooperka , thanks for reviewing the PRs.

I also believe that customizations should be left to the user and not to the core component but I also think that showing the name of the user sending a chat message should be a core capability of the component. Just showing the avatar or the initials is not enough because two more users may have same initials. This is something that many users may want and shouldn't need to define a custom `renderBubble` method.

kfiroo:
@varungupta85 I'm with @cooperka here.
There are already far too many customizations to the chat, and I would be happy to remove some of them, not add more :)
I think the better solution would be to provide users with easy ways to customize their components instead of hiding it behind high level props with obscure meaning.

varungupta85:
@kfiroo Ability to show the name of the sender of the message seems like a very core capability to me for a chat component or else you are receiving messages but you don't know who is sending them. But this is just my opinion. Please let me know if you want me to close the PR.

bexoss:
I need it... showing username is core of chat.

cooperka:
Hi @varungupta85 and @bexoss, the way to do this right now is to override `renderBubble` and add the name yourself. It should be straightforward without needing any change to the library code. If you're having trouble please post here and I can write up an example. Thanks!

bexoss:
screen shot 2017-09-07 at 11 17 33 pm

Now I applied like this. I think this is more natural...
I edited `Bubble.js`

render() {

    let username = null;
    if(this.props.showUsername &&
        (!isSameUser(this.props.currentMessage, this.props.previousMessage) ||
        !isSameDay(this.props.currentMessage, this.props.previousMessage)) &&
        (this.props.currentMessage.user._id !== this.props.user._id)
    ){
        username = (
            <View style={this.props.usernameContainerStyle}>
              <Text style={this.props.usernameStyle}>{this.props.currentMessage.user.name}</Text>
            </View>
        )
      }

    return (
      <View style={[styles[this.props.position].container, this.props.containerStyle[this.props.position]]}>
        {username}
       ....

mafiusu:
Hey @cooperka can you please write an example of how to display a username? I am trying but it doesn't fit very well with the position.

cooperka:
Hi @mafiusu, I just opened a PR to show a full example -- see the code here: https://github.com/FaridSafi/react-native-gifted-chat/pull/649

bexoss:
Hi @mafiusu , If you need a example please check my forked repository of gifted-chat, there is a example I think, I don't remember where is it exactly, but you can find it easily.

ruk91:
Can I implement this on default example?

dengue8830:
my very basic solution. Im still trying to figure it out how to show the username just for the last user's chat group

```
    renderBubble={_props => 
          <View>
            <Bubble
              {..._props}
              .........
            />
            <Text>{_props.currentMessage.user.name}</Text>
          </View>
    }
```
image

G3z:
@dengue8830 you can use `isSameUser` to determine if you need to show the name

I built this from your example

```javascript
renderBubble={props => {
	let name;
	if (!props.isSameUser(props.currentMessage, props.previousMessage)) {
		name = (
			<Text style={styles.username}>
				{props.currentMessage.user.name}
			</Text>
		);
	}

	return (
		<View>
			{name}
			<Bubble {...props} />
		</View>
	);
}}
```

G3z:
I don't know if it's the best way but you have the control you need so why not ?

dengue8830:
yes its great, thanks!!

PR #450, opened 2017-05-02
Fix wrong styling props used for some components

varungupta85:
Some of the internal components are using `containerStyle` styling prop to allow the user to change the way those components from within their app but the `containerStyle` prop is filtered out when the `Bubble` is rendered. It is done in `Message.js` in the `getInnerComponentProps` method

```javascript
  getInnerComponentProps() {
    const {containerStyle, ...props} = this.props;
    return {
      ...props,
      isSameUser,
      isSameDay
    }
  }
```

Due to this, `containerStyle` prop is never transferred to components such as `Avatar`, `Time` which are using `containerStyle` to change their styling.

I have renamed the `containerStyle` prop in such components such that their styling is configurable from within the app.

I have also renamed the `textStyle` prop name to `timeTextStyle` which could be a breaking change but I couldn't understand under what circumstances would I want to use same styles for message text and time. I would be happy to merge it with `textStyle` to make it backward compatible if maintainers think that is a better solution. I feel that if I do merge them together, then I would have to explicitly cancel out some styles defined in `textStyle` if I don't want to use them for time text.

If I have misunderstood the code, please let me know.

### DIFF
--- a/src/Avatar.js
+++ b/src/Avatar.js
@@ -79,7 +79,7 @@ Avatar.defaultProps = {
     user: null,
   },
   nextMessage: {},
-  containerStyle: {},
+  avatarContainerStyle: {},
   imageStyle: {},
   //TODO: remove in next major release
   isSameDay: warnDeprecated(isSameDay),
@@ -92,7 +92,7 @@ Avatar.propTypes = {
   currentMessage: React.PropTypes.object,
   nextMessage: React.PropTypes.object,
   onPressAvatar: React.PropTypes.func,
-  containerStyle: React.PropTypes.shape({
+  avatarContainerStyle: React.PropTypes.shape({
     left: View.propTypes.style,
     right: View.propTypes.style,
   }),
--- a/src/Avatar.js
+++ b/src/Avatar.js
@@ -25,7 +25,7 @@ export default class Avatar extends React.Component {
 
     if (isSameUser(this.props.currentMessage, messageToCompare) && isSameDay(this.props.currentMessage, messageToCompare)) {
       return (
-        <View style={[styles[this.props.position].container, this.props.containerStyle[this.props.position]]}>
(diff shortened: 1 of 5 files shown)


varungupta85:
hi @cooperka, please review this PR also if you have some time. Either I have completely misunderstood how the custom styles are applied or this is a huge oversight in the code. I have tested the code and I was able to apply custom styles to many components only after I made these changes. I will update the README if you agree with the changes.

kfiroo:
@varungupta85 As we discussed in the user name thread - this is what I want to avoid.
When we have too many high level customizations these kind of confusion happens.

I think customizations of inner components should only be possible be overriding this component and using its props (or your own custom component).
For example, say you want to change some style of the bubble, I wouldn't want to provide `containerStyle` to `GiftedChat` as the name is confusing and might be used by other components (exactly what happened here).
Also, having 6 different `xxxContainerStyle` is not a good option in my opinion as it clutters the API.
Component props are like the parameters of a constructor, and you wouldn't provide your ctor with 25 different params would you? :)

I would, however, consider moving all the customizations into the inner components so `Bubble` could have its own set of props that are used to customize it, and the user could use `renderBubble` and simply override those props. (if that's not enough you could always render your own comp here)

varungupta85:
My intent was to fix the current styling more than introducing new props and I understand that in the process I have introduced new props because I just thought that was how it was supposed to be because the `containerStyle` prop is filtered out while rendering the `Bubble`, so that the `containerStyle` prop wouldn't be passed down to any of these sub-components. I guess I didn't realize that the `containerStyle` prop may be passed down when these components are used directly by the user.

I feel that the way the styling is setup currently is a bit confusing for a new user but I understand it a little better now.

I will close the PR.

kfiroo:
@varungupta85 Thanks for understanding!
I completely agree with you, the way styles are currently handled is far from ideal.
I hope I will find some time soon to fix that, my greatest fear is that the change is going to break all existing installations. :/

PR #453, opened 2017-05-04
fix isSameDay function

varungupta85:
`isSameDay` method was returning `true` if we feed a `currentMessage` which is created today and an empty `diffMessage`. That is because `moment(undefined)` call will initialize the moment object to today's date. I think we were using `isValid` to make sure that there is a valid `createdAt` for `currentMessage` and `diffMessage` but `moment(undefined).isValid()` will return true.

I have changed the method to make sure that we have a valid `createdAt` date for `diffMessage` and remove the `isValid()` calls as they will always return true.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -4,12 +4,16 @@ const DEPRECATION_MESSAGE = 'isSameUser and isSameDay should be imported from th
 
 export function isSameDay(currentMessage = {}, diffMessage = {}) {
 
-  if(!diffMessage.createdAt) {
+  if (!diffMessage.createdAt) {
     return false
   }
 
   let currentCreatedAt = moment(currentMessage.createdAt);
   let diffCreatedAt = moment(diffMessage.createdAt);
+
+  if (!currentCreatedAt.isValid() || !diffCreatedAt.isValid()) {
+    return false;
+  }
 
   return currentCreatedAt.isSame(diffCreatedAt, 'day');
 
--- a/src/utils.js
+++ b/src/utils.js
@@ -4,12 +4,12 @@ const DEPRECATION_MESSAGE = 'isSameUser and isSameDay should be imported from th
 
 export function isSameDay(currentMessage = {}, diffMessage = {}) {
 
+  if(!diffMessage.createdAt) {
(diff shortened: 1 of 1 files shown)


src/utils.js
```diff
@@ -4,13 +4,13 @@ const DEPRECATION_MESSAGE = 'isSameUser and isSameDay should be imported from th
...
 export function isSameDay(currentMessage = {}, diffMessage = {}) {
 
+  if(!diffMessage.createdAt) {
+    return false
+  }
+
   let currentCreatedAt = moment(currentMessage.createdAt);
   let diffCreatedAt = moment(diffMessage.createdAt);
 
-  if (!currentCreatedAt.isValid() || !diffCreatedAt.isValid()) {
```
cooperka: Why have both of these checks been removed?

src/utils.js
```diff
@@ -4,13 +4,13 @@ const DEPRECATION_MESSAGE = 'isSameUser and isSameDay should be imported from th
 
 export function isSameDay(currentMessage = {}, diffMessage = {}) {
 
+  if(!diffMessage.createdAt) {
```
cooperka: Nit: space after `if`

src/utils.js
varungupta85: I found that `isValid` returns true for all the cases that this test may be trying to avoid. For e.g. `moment(undefined).isValid()`, `moment(0).isValid()`, `moment({}).isValid()` will all return true. I guess one case where it will return false is for `moment(null).isValid()`. I will put the check back in.

varungupta85:
@cooperka I have made the changes you requested. I am not sure how I can help with the testing. I have tested it on my end but I guess it would be good to know that this change is not breaking anything expected for some other users.

kfiroo:
@cooperka Hey, You are absolutely right, we must have some tests here.. I promised myself I'd find time for it but couldn't get to it :(

I think the hardest part is to set it up, from there on we could all just add tests as we add/change functionality.

If anyone has some time to set up some basic tests it would be amazing, otherwise I'll get to it at some point :)

Issue #423, opened 2017-04-13
[jest] Identifier 'exports' has already been declared


Maxwell2022:
#### Issue Description
Jest tests is throwing error in projects as soon as you import `react-native-gifted-chat`.
This is related to https://github.com/expo/react-native-action-sheet/issues/26

`react-native-gifted-chat` is using an outdated version (`0.3.0`) of `@expo/react-native-action-sheet` (latest: `0.3.1`)

#### Steps to Reproduce / Code Snippets

Add a jest test file in your project that is testing a basic component. Add an import to this component: 
`import Chat from 'react-native-gifted-chat';`

#### Expected Results
No errors from jest

#### Additional Information

* React Native version: 0.42.3
* react-native-gifted-chat version: ^0.1.4
* Platform(s) (iOS, Android, or both?): any

![image](https://cloud.githubusercontent.com/assets/670701/24986548/38539eb6-203e-11e7-9bb3-2e82a4fdbf02.png)

kfiroo:
@Maxwell2022 Thanks! want to create a quick PR to fix that? :)

brunsy:
created the PR for this: https://github.com/FaridSafi/react-native-gifted-chat/pull/454

bitsal:
When is it going to be released?

Maxwell2022:
PR has been merged, I'm closing this issue. No idea about the release @bitsal

PR #454, opened 2017-05-04
Update @expo/react-native-action-sheet

brunsy:
Update @expo/react-native-action-sheet,
previous versions causes issues running jest tests.

https://github.com/FaridSafi/react-native-gifted-chat/issues/423


RWOverdijk:
👍 when is this being released?

PR #464, opened 2017-05-18
[EarlierButton] if not necessary hide button

sqlProvider:
Check default page size and if not necessary hide button.


omeraplak:
@FaridSafi can you merge this?

PR #467, opened 2017-05-26
flex instead of full width/height for images.

stokesbga:
Setting image to full width/height makes animation returning image to previous spot look extremely awkward. flex: 1 allows image to resize between thumbnail and fullscreen as was intended with lightbox.


cooperka:
Hey, this is wonderful! I had the same problem in my own app, and this solves it. Thank you @stokesbga.

PR #475, opened 2017-06-02
increasing the max input so i can render custom input view

nyo16:
This is a temporary fix so i be able render custom view for input.
I can prepare a full fix (aka setting MAX_COMPOSER_HEIGHT from props etc.) if more ppl are interested in.


nyo16:
@FaridSafi do you have something in your mind of how to handle this? 
I wrote some code also to convert the InputToolbar in a more generic input container. 
Are you interested in to add this as a new feature?

Thank you for this amazing lib :)

PR #479, opened 2017-06-08
Fix default rendering of avatars

cooperka:
Simple fix: default `renderAvatar` to `undefined` instead of `null`, now that `null` causes it to not be rendered at all.

Originally mentioned in https://github.com/FaridSafi/react-native-gifted-chat/pull/447#issuecomment-305815371; see there for more context.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 node_modules/
 npm-debug.log
 TODO.md
+
+.idea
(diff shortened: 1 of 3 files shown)


cooperka:
@henrythach can you please test this and confirm it solves the problem you mentioned?

cooperka:
@kfiroo any chance you can merge soon? Pretty big bug in default behavior.

PR #484, opened 2017-06-14
add renderContainerStyle props on root GiftedChat

Nours312:
just to style root container, maybe I'm not the once one who need that.


dani-sc:
Could really make use of this feature, so I can change the background color.
@FaridSafi, any changes needed to merge this PR?

Kureev:
Hey folks! I came along having exactly the same issue you're trying to solve. After a brief dive into the codebase, it seems there is a pretty straightforward way of achieving this: 

```js
<GiftedChat
  listViewProps={{
    style: {
      backgroundColor: '#fff',
    },
  }}
  ...
/>
```

PR #478, opened 2017-06-07
Change deprecated View.propTypes to ViewPropTypes

mgustin12:
[477](https://github.com/FaridSafi/react-native-gifted-chat/issues/477)

PR #488, opened 2017-06-21
Replace View.propTypes with ViewPropTypes

mitchellbutler:
Fix for React Native’s Warning: “View.propTypes has been deprecated and will be removed in a future version of ReactNative. Use ViewPropTypes instead”


mitchellbutler:
Removing this because I missed #478, an open PR with the same fix.

PR #504, opened 2017-07-11
Take into account the keyboard height in onLayout

grin:
Without this dynamic layout changes occuring while keyboard is on can result
in pushing the input field behind the keyboard and making it invisible.


cooperka:
Hi @grin, the change makes sense to me, but can you please give some steps to reproduce the original issue? That will help me verify your change.

grin:
@cooperka I ended up making a different fix: https://github.com/grin/react-native-gifted-chat/commit/4e731c87b2fd51bc4444b2673b31e50cf28dadff

I need this fix for when GiftedChat is added to the DOM tree when keyboard is already visible. That's the only way I could initialize the chat component with the correct height.

PR #508, opened 2017-07-14
Migrate to prop-types library

wsabransky:
`React.PropTypes` is deprecated as of React v15.5. This migrates to the `prop-types` library which should have no functional consequences.


cooperka:
Thank you @wsabransky!

Issue #495, opened 2017-06-30
Avatar for user is not working


kwnccc:
When passing down messages based on Message Object defined in the documentation, avatar cannot be set for user. The same message object works fine for bot avatar, but I'd like to be able to add an avatar also for the user.

Example bot message object: 
```
{
  _id: generateMessageID(),
  text: talk.response,
  createdAt: new Date(),
  user: {
    _id: 2,
    name: botName,
    avatar: botAvatar
  },
  metadata: talk,
}
```

Example bot message object: 
```
{
  _id: generateMessageID(),
  text: data.response_text,
  createdAt: new Date(),
  user: {
    _id: 1,
    name: userName,
    avatar: userAvatar
  },
  metadata: data
}
```

* React Native version: 0.45.1
* react-native-gifted-chat version: 0.1.4

kwnccc:
After a fast digging into the code I assume this (extra?) check is causing second avatar not to appear. Is this check happening on purpose? Git history shows that it mostly is because of some backwards compatibility. Is this still the case? If not I can open a PR about it.

https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/Message.js#L44

jeaye:
I'd also like to show an avatar for the user; @kwnccc did you end up making a change which you could turn into a PR? Thanks!

kwnccc:
@jeaye not yet, I was hoping for a response from maintainers, in case my assumption is not correct. I can open a PR and further discuss the proposed change there, I'm not sure how it will impact the rest of the implementation.

cooperka:
Hi @kwnccc, please try again on latest. This was likely fixed in https://github.com/FaridSafi/react-native-gifted-chat/pull/479.

cooperka:
Closing for now, please let me know if you're still having issues.

kwnccc:
@cooperka after a trying out the latest version `0.2.2` this issue doesn't seems resolved unfortunately. I'm trying with 2 different ways, either initialising user object on GiftedChat component with an avatar or adding it as a parameter to the messages user object, but none of this worked. 

To better illustrate the needed result I'm attaching 2 screenshots, the first one with the current behaviour, and the second with the needed one.
[](http://imgur.com/a/GIgTQ)

cooperka:
[...]

PR #511, opened 2017-07-21
Add 'showUserAvatar' prop; don't render null avatars

kwnccc:
- check if current message user has an avatar to display
- don't base the logic of the display of the avatar on the user object id

Issue: #495
Issue URL: https://github.com/FaridSafi/react-native-gifted-chat/issues/495

Screenshot with the behaviour before:
screen shot 2017-07-21 at 14 48 27

Screenshot with the behaviour after:
screen shot 2017-07-21 at 14 50 00

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ e.g.
 - **`renderLoading`** _(Function)_ - Render a loading view when initializing
 - **`renderLoadEarlier`** _(Function)_ - Custom "Load earlier messages" button
 - **`renderAvatar`** _(Function)_ - Custom message avatar; set to `null` to not render any avatar for the message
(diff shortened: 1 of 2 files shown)


kwnccc:
Relative issue #495

kwnccc:
@cooperka Please give a look and let know what you think :)

cooperka:
Thank you @kwnccc. I don't have time to test it out right now, but the change looks good to me, and I'll try to verify this weekend if possible.

cooperka:
@kwnccc there were a few problems with the original implementation; I added a couple more commits including a `showUserAvatar` prop, and I think it's much better now. Thanks again for the help!

jeaye:
Awesome, thanks guys!

kwnccc:
@cooperka Awesome thanks on the fast response and review!

jasonFullDev:
it's not a bug ... it's a feature

PR #550, opened 2017-08-25
Fix wrong order of actions in onPhonePress()

ianlin:
The order of the actions index should be the same as the options of action sheet.


cooperka:
Fixed in https://github.com/FaridSafi/react-native-gifted-chat/pull/521, thanks for your fix as well :)

PR #551, opened 2017-08-25
Add parsePatterns prop

ianlin:
Add a custom property `onPhonePress()` to deal with the action.

P.s. code based on PR #550

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,15 +150,16 @@ e.g.
 - **`keyboardShouldPersistTaps`** _(Enum)_ - Determines whether the keyboard should stay visible after a tap; see [`<ScrollView>`](https://facebook.github.io/react-native/docs/scrollview.html) docs
 - **`onInputTextChanged`** _(Function)_ - Callback when the input text changes
 - **`maxInputLength`** _(Integer)_ - Max message composer TextInput length
-- **`parsePatterns`** _(Function)_ - Custom parsed patterns, e.g.
-  ```javascript
-  <GiftedChat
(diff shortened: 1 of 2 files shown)


cooperka:
Thanks for this @ianlin! Would you mind rebasing on master and also making it a bit more generic? That way we could handle every type of link as well as [custom regex links](https://github.com/taskrabbit/react-native-parsed-text). I think rather than `onPhonePress` we can add a prop called `parsePatterns`, e.g.

```jsx
const linkStyle = StyleSheet.flatten([styles[this.props.position].link, this.props.linkStyle[this.props.position]]);

<ParsedText
  style={[styles[this.props.position].text, this.props.textStyle[this.props.position]]}
  parse={[
    ...this.props.parsePatterns(linkStyle),
    {type: 'url', style: linkStyle, onPress: this.onUrlPress},
    {type: 'phone', style: linkStyle, onPress: this.onPhonePress},
    {type: 'email', style: linkStyle, onPress: this.onEmailPress},
  ]}
>
```

and for your specific use-case you'd pass in

```jsx
<GiftedChat
  parsePatterns={(linkStyle) => [
    { type: 'phone', style: linkStyle, onPress: this.customOnPhonePress },
  ]}
/>
```

What do you think?

ianlin:
Thanks @cooperka that is totally make sense. Have updated the PR, cheers!

cooperka:
Merging but I don't have time right now to fully test before release. I'll release this later (likely several days from now).

PR #523, opened 2017-07-27
Updated dependencies for compatibility with React 16 and RN > 0.46

brunocascio:
Since React 16 use `prop-types` as an external package, in order to use `PropTypes`, I've just added this in my fork.

I've made PRs on the plugins as well: 

https://github.com/taskrabbit/react-native-parsed-text/pull/34 (Merged)

https://github.com/expo/react-native-invertible-scroll-view/pull/49 (Merged)

https://github.com/oblador/react-native-lightbox/pull/75 (PR of @formula1) (Merged)

PR #552, opened 2017-08-28
Update package.json

bintoll:
The react-native-invertible-scroll-view in dependencies should be 1.1.0 because this version has a fix for PropTypes for  react-native >=0.44. Also the react-native-lightbox and react-native-parsed-text should be fixed. For example here: https://github.com/bintoll/react-native-lightbox and https://github.com/brunocascio/react-native-parsed-text

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-native-gifted-chat",
-  "version": "0.2.5",
+  "version": "0.2.7",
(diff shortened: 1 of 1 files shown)


brunocascio:
Should be fixed in https://github.com/FaridSafi/react-native-gifted-chat/pull/523.

Please, feel free to reopen.

Thanks!

PR #556, opened 2017-08-30
Update invertible-scroll-view dependency to 1.1.0

lesserj:
The 1.1.0 version addresses the issue in #523 for the react-native-invertible-scroll-view library.


lesserj:
Nevermind.  @bintoll beat me to it with #552.

PR #563, opened 2017-09-06
Animation on input size

itinance:
This PR reflects also the "isAnimated"-Property when the composer increases its height e.g. when long text with multiple lines is entered or if someone wants to increase the composer-Size programmatically with onInputSizeChanged()

It will not break any compatibility.


itinance:
i've closed it because it already contains also another PR from me. I will create  a new one

PR #573, opened 2017-09-14
Fixed doc link for long press

nonotest:
the link goes to a custom action file which has no reference to onLongPress.
(the example is for the input bar)

at least link to the source file which has the default behaviour eg copy text?

cheers,


cooperka:
Thanks @nonotest.

PR #595, opened 2017-10-02
Fix input toolbar position

jjdp:
This positions the toolbar to be at the absolute bottom since I scale my gifted chat depending on the devices


brunocascio:
Thanks for contributing!

PR #605, opened 2017-10-12
Replace moment-with-locales.min with regular moment

doomsower:
After upgrade to `react-native@0.49.3` and `moment@2.19.1` I got error:

```
moment-with-locales.min.js require() must have a single string literal argument
```

Turns out `react-native-gifted-chat` explicitly import from 'moment/min/moment-with-locales.min';
Removing this fixes my build. In my project I use non-english locale and everything seems to be ok after switch to original `moment`.

If `moment/min/moment-with-locales.min` is imported on purpose please close this PR.

### DIFF
--- a/src/Day.js
+++ b/src/Day.js
@@ -7,7 +7,7 @@ import {
   ViewPropTypes,
 } from 'react-native';
 
-import moment from 'moment/min/moment-with-locales.min';
+import moment from 'moment';
 
 import { isSameDay, isSameUser, warnDeprecated } from './utils';
 
--- a/src/GiftedChat.js
+++ b/src/GiftedChat.js
@@ -8,7 +8,7 @@ import {
 } from 'react-native';
 
 import ActionSheet from '@expo/react-native-action-sheet';
-import moment from 'moment/min/moment-with-locales.min';
+import moment from 'moment';
 import uuid from 'uuid';
 
 import * as utils from './utils';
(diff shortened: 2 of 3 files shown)


jspiers:
@doomsower Yes! Thank you for this. With this fix I'm finally running react-native-gifted-chat under RN 0.49.

tonyhb:
I'd actually recommend ditching moment in favor of date-fns: https://date-fns.org/

date-fns allows es6 imports of individual functions, so we'll have a smaller bundle size because we only import what we need (as opposed to the entirety of momentjs).

brunocascio:
I'm going to try this now. If it works, I will merge it 💃

kelset:
Any update on this getting merge? Even on the metro-bundler issue it looks like this PR is the only way to resolve the issue

brunocascio:
Hi guys, sorry for the delay. I cannot be able to tried RN 0.49 since I updated to iOS 11 (I got a segmentation fault error). However, I tested in my current project using `import moment from 'moment'` and it works properly.

By the way, since moment has a lot of problems wit react-native ecosystem, a PR with the migration to date-fns would be appreciated.

Thanks to everyone!

brunocascio:
Published `v0.2.9` 💃

tonyhb:
Thanks for the heads up bruno - if I get time at work I'll take a look at migrating to date-fns.

PR #629, opened 2017-11-06
Modified for FlatList

gavin-gmlab:
Implementation for FlatList instead of ListView component


src/MessageContainer.js
```diff
@@ -67,10 +51,9 @@ export default class MessageContainer extends React.Component {
     if (this.props.messages === nextProps.messages) {
       return;
     }
-    const messagesData = this.prepareMessages(nextProps.messages);
     this.setState({
-      dataSource: this.state.dataSource.cloneWithRows(messagesData.blob, messagesData.keys)
-    });
+      messagesData: this.prepareMessages(nextProps.messages)
```
sondremare: This will cause all chat messages to re-render every time a new message is added.

redwind: i think you should remove prepareMessages and move set value for nextMessage and previousMessage into renderItem function

gavin-gmlab: My thoughts:
Its not worse than what it was with the ListView approach. Correct me if I am wrong since the old approach was rerendering the whole ListView anyway. FlatList is supposed to render only what is actually viewable anyway, and not those components that are outside the view, so we are getting an improvement right there.

I was more thinking of getting a quick fix in for now, and we can always have perf improvements later? But just my opinion.

wizza-smile:
maybe you guys have some thoughts to share on this! @Kabangi @radko93 @dralletje @bartolkaruza @DavidKuennen @Ashoat

src/MessageContainer.js
dralletje: It is not the prepareMessages that will cause everything (in the view) to rerender, that will happen anyway when a new message arrives. What we can do about that is make the components being rendered pure, and allow them to be effective PureComponent-s

src/MessageContainer.js
```diff
@@ -20,37 +19,22 @@ export default class MessageContainer extends React.Component {
...
-          nextMessage,
-          hash: md5(toHash)
-        };
-        return o;
-      }, {})
-    };
+    return output = messages.reduce((o,m,i) => {
+      const previousMessage = messages[i + 1] || {}
+      const nextMessage = messages[i - 1] || {}
+      o.push( {
```
dralletje: Two small questions here:

1. It seems to me this is a lot simpler done using `.map`, maybe even more performant? (Dunno if that even matters here)

2. Because you **merge** previousMessage and nextMessage props onto the existing message, that message will be "invalidated" when matching it against existing messages (causing PureComponent or shouldComponentUpdate to always see change).

src/MessageContainer.js
```diff
@@ -99,66 +82,68 @@ export default class MessageContainer extends React.Component {
...
-      position: message.user._id === this.props.user._id ? 'right' : 'left',
+      key: item._id,
+      currentMessage: item,
+      previousMessage: item.previousMessage,
+      nextMessage: item.nextMessage,
+      position: item.user._id === this.props.user._id ? 'right' : 'left',
     };
 
     if (this.props.renderMessage) {
       return this.props.renderMessage(messageProps);
```
dralletje: Is it on purpose that this is outside of the `{ transform: [{ scaleY: -1 }] }` ? I don't think people will want to have to set that themselves, and when we go to use `invert={true}` on the container later, the people that do have their own `{ transform: [{ scaleY: -1 }] }` will have to remove it again and ughhhhh hahaha :)

src/MessageContainer.js
dralletje: About the second one, I see that Message isn't PureComponent yet, but I think it is still good to these optimisations already, if we want Message to be an effective PureComponent later ;)

https://github.com/gavin-gmlab/react-native-gifted-chat/blob/33dc31f41d29be98e441c4b27ad74e453b9801b5/src/Message.js#L16

src/MessageContainer.js
```diff
@@ -99,66 +82,68 @@ export default class MessageContainer extends React.Component {
...
-          dataSource={this.state.dataSource}
+          data={this.state.messagesData}
 
-          renderRow={this.renderRow}
+          renderItem={this.renderRow}
           renderHeader={this.renderFooter}
-          renderFooter={this.renderLoadEarlier}
-          renderScrollComponent={this.renderScrollComponent}
+          renderFooter={this.renderLoadEarlier()}
+          style={{transform: [{scaleY: -1 },{perspective: 1280}]}}
```
dralletje: One last thing, what is the reasoning behind using this transform ourselves, and not using `inverted={true}`? I'm not sure why it not in the documentation (so you may very well know something I don't) :-)

sondremare: Inverted works only on RN 0.49 and above. Atleast for Huawei phones running Android 7

dralletje: Then maybe it is even a combination of the fact that inverted got added later + the perspective fix... Maybe good to wait some version before we use `inverted`, but still worth adding a TODO/NOTE?

xcarpentier:
Hello, thanks !
is it possible to merge master into this branch to have lint and jest ?

xcarpentier:
I've been creating https://github.com/FaridSafi/react-native-gifted-chat/pull/705 to have a clean history.

Please let me know if I can close this one?

gavin-gmlab:
@xcarpentier , yes, feel free to close this PR

PR #637, opened 2017-11-15
OnPress event on Bubble

gouravd:
This adds an onPress event on the Bubble because even with wrapping the Bubble with TouchableNoHighlight, the tap,press event on Bubble never bubbles up to the TouchableNoHighlight

A typical use case is show options like Like, Share etc...which is not appropriate for LongPress

### DIFF
--- a/src/Bubble.js
+++ b/src/Bubble.js
@@ -239,7 +239,7 @@ Bubble.defaultProps = {
 Bubble.propTypes = {
   touchableProps: PropTypes.object,
   onLongPress: PropTypes.func,
-  onPress: PropTypes.fubc,
+  onPress: PropTypes.func,
   renderMessageImage: PropTypes.func,
   renderMessageText: PropTypes.func,
   renderCustomView: PropTypes.func,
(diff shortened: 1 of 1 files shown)


pavanmehta91:
Please merge this. It's useful

michael-gillett:
Agreed this would be great to have

Jhuff6:
We are going to have to have the same thing as the other thing that you
have the right

peterkappelt:
In case someone is coming from Google, like me, and is yet trying to figure out how to do it:

The pull request hasn't been accepted, apparently. There is no onPress property in the current release.
The solution (for now and for my use case) is a custom bubble renderer, something like this:
```
const customBubbleRenderer = props => {
    const internal = props.currentMessage.text.internal;
  
    return (
      <Bubble
        {...props}
        touchableProps={{
          onPress: () => console.log("Bubble has been pressed"),
        }}
        textStyle={{
          ...
        }}
   
      />
    );
  };
```

and then obviously:
```
<GiftedChat
        renderBubble={customBubbleRenderer}
        ....
      />
```

PR #643, opened 2017-11-21
Remove redundant ref causing crash in React 16.1

oblador:
With React 16.1.1 this component causes a crash due to string refs being deprecated. Couldn't see this ref actually being used anywhere so I removed it.


cooperka:
Thank you @oblador! Note to publisher: we should merge, but this will be a breaking change for anyone using this exported component.

PR #649, opened 2017-11-27
Add example "Slack" UI

cooperka:
👀 See the screenshots [here](https://github.com/FaridSafi/react-native-gifted-chat/blob/92ce5963a9857213e53751fc2f9fa654fc6dcde9/example-slack-message/README.md).

### Motivation

- It's nice to have a full example of overriding the UI
- Many people want to display usernames
- Many people want to display all messages on the same side

### Changes

- Added 3 new JS files + screenshots under `example-slack-message/`
- Updated the README to point to these examples

### Notes

- The new README points to `master` for the comparison images, where the screenshots don't exist yet; use [the previous commit](https://github.com/FaridSafi/react-native-gifted-chat/tree/example-slack-message^/example-slack-message) to see what it would really look like


phillip055:
Really love this

xcarpentier:
Why not after merging this, add an option to use it in this lib ? Not just as an example.

cooperka:
@xcarpentier that would be great -- feel free to add support for that! I don't have the time right now myself, but I figured I should at least submit this example from my app.

cooperka:
Though, maybe we could add it just to the example app, so the example works live? It won't be quite as effective as an example if it's simply turned into a feature.

xcarpentier:
Why not using a expo app for demo like this? It will be more easy to see and test.
I think I will add soon a travis build to deploy example on expo...

cooperka:
The [current example app](https://github.com/FaridSafi/react-native-gifted-chat/tree/master/example) doesn't use expo, but that would certainly be easy to integrate. Thanks!

xcarpentier:
LGTM 👌

PR #489, opened 2017-06-22
Start messages at top

magfurulabeer:
This is a simple fix. It just turns off the inverted prop of the inverted scroll view and has new messages appended to current messages rather than the other way around.

PR #658, opened 2017-12-09
Inverted prop for reverse messages display order

apparition47:
Amendment to the #489 PR allow the user to set a prop to reverse the `messages` display order on the scroll view.

 **`inverted`** _(Bool)_ - Reverses display order of `messages`; default is `true`

### DIFF
--- a/src/GiftedChat.js
+++ b/src/GiftedChat.js
@@ -84,14 +84,14 @@ class GiftedChat extends React.Component {
     if (!Array.isArray(messages)) {
       messages = [messages];
     }
-    return this.inverted ? messages.concat(currentMessages) : currentMessages.concat(messages);
+    return this.props.inverted ? messages.concat(currentMessages) : currentMessages.concat(messages);
   }
 
   static prepend(currentMessages = [], messages) {
     if (!Array.isArray(messages)) {
       messages = [messages];
     }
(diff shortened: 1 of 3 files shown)


xcarpentier:
@brunocascio why you don't merge this one?

src/GiftedChat.js
```diff
@@ -84,14 +84,14 @@ class GiftedChat extends React.Component {
     if (!Array.isArray(messages)) {
       messages = [messages];
     }
-    return currentMessages.concat(messages);
+    return this.inverted ? messages.concat(currentMessages) : currentMessages.concat(messages);
```
xcarpentier: Should be `this.props.inverted`
Please fix.

xcarpentier:
@brunocascio I've made the change in PR: `this.inverted => this.props.inverted`

brunocascio:
Hi @xcarpentier ! Thanks for the PR. Merging

xcarpentier:
This is not good at all, I hope that works when I fixed but I the original PR is not form me but from @apparition47 ... and now in my test in master it's not so good :/

apparition47:
@xcarpentier looks like you went ahead and added those changes [here](https://github.com/FaridSafi/react-native-gifted-chat/pull/658/commits/528aa0e6abc1d17b1500203003e02c9520d830cb)

PR #677, opened 2017-12-24
Unnecessary semicolon removed

devshekhawat:
This file was ending with 2 semicolons, this produces Unnecessary semicolon error. By this update the file is error free.


xcarpentier:
Thanks

Issue #611, opened 2017-10-21
Render Avatar on each message


tangweikun:
#### Issue Description

I want to show avatar on each message, not only show on the top or bottom, like the screenshot.
I just fork your repo and comment some code to make it work as my way.
Can you add a new feature to make that work?
Thanks!

![image](https://user-images.githubusercontent.com/13361639/31847700-154b2e14-b5e6-11e7-9376-7b008cfe2843.png)

juanmanuelarze:
In MessageContainer.js line 42 you have a comment 
// add next and previous messages to hash to ensure updates
you just need change the program for "not ensure updates"

Abhishekgarg727:
@tangweikun  i know its  late but for any other user who having issue to show avatar on every chat message , i think this will help.

**Solution
So i use two props of gifted chat:**

```
 showUserAvatar={true}
 showAvatarForEveryMessage={true}
```

**my code snippet**
```
<GiftedChat
                    messages={this.state.messages}
                    onSend={messages => {this.onSend(messages)}}
                    user={{_id: 1}} 
                    maxInputLength={1000}
                    alwaysShowSend={true}
                    renderBubble={(props)=>this.renderBubble(props)}
                    renderActions={(props) => this.renderActions(props)}
                    showUserAvatar={true}
                    showAvatarForEveryMessage={true}
                    parsePatterns={linkStyle => [
                            {
                                pattern: /#(\w+)/,
                                style: { ...linkStyle, color: 'lightgreen' },
                                onPress: props => alert(`link`),
                            },
                    ]}
                />
```

hope this helps.

PR #729, opened 2018-01-25
Add Prop to always show avatars for each message

### LINKED ISSUE #507: How to show avatars in every messages?
Hi, I want to render avatars in every messages. Any ideas?

apparition47:
Resolves #507 #611 

Current behaviour only shows avatars if the consecutive message's timestamp and user is different. This will allow the user to toggle avatars to always show. Defaults to `false` to maintain existing behaviour.

Usage:
```react
   <GiftedChat
          showAvatarForEveryMessage={true}
        />
```


apparition47:
I just realized there's already an existing PR #297 that has the exact same changes.

PR #756, opened 2018-02-17
Ts support

ascariandrea:
I've spotted some unused props during the type definitions that can be safely removed, but I would do it in anothe PR.

Let me know if these types are ok :)


ascariandrea:
Are you planning to merge within a few days?

brunocascio:
LGTM

Issue #761, opened 2018-02-20
iPad orientation change -> Text Input component's UI glitch - Cut off 


sujameslin:
#### Issue Description
In iPad, when it goes to landscape mode from portrait mode, the message input box UI went cut off. 
Please review attached expo link to reproduce.
![screen shot 2018-02-23 at 10 35 39 pm](https://user-images.githubusercontent.com/10737745/36599466-6b0c57c6-18ea-11e8-861e-89c838972efe.png)

#### Steps to Reproduce / Code Snippets
https://exp.host/@sujameslin/example-expo

#### Expected Results
No matter what orientation mode the ipad is, it should fully render the message input box without any glitch.

#### Additional Information

* Nodejs version: 9.1.0
* React version: 16.0.0
* React Native version: Expo React Native SDK 24.0.0
* react-native-gifted-chat version: 0.3.0
* Platform(s) (iOS, Android, or both?): tested only iOS

PR #764, opened 2018-02-23
set InputToolBar to full width of screen dynamically

sujameslin:
This addresses #761

### DIFF
--- a/src/InputToolbar.js
+++ b/src/InputToolbar.js
@@ -82,7 +82,10 @@ export default class InputToolbar extends React.Component {
   render() {
     return (
       <View
-        style={[styles.container, this.props.containerStyle, { position: this.state.position }]}
+        style={[styles.container, this.props.containerStyle, {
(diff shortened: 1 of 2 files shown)


joel1st:
Looks good to me!

src/InputToolbar.js
```diff
@@ -82,7 +82,10 @@ export default class InputToolbar extends React.Component {
   render() {
     return (
       <View
-        style={[styles.container, this.props.containerStyle, { position: this.state.position }]}
+        style={[styles.container, this.props.containerStyle, {
+          position: this.state.position,
+          width: Dimensions.get('window').width,
```
gianpaj: wouldn't this calculate the width every time the InputTool rendered? I guess is better to calculate it when the component willMount and when the orientation changes

sujameslin:
There's better approach - #776 
Hence i'm closing this.

PR #765, opened 2018-02-25
Add moment as peerDependency

### LINKED ISSUE #762: Locale prop no longer working
#### Issue Description

The prop `locale` doesn't seem to work anymore for the time separators in the chat. 

You find it here: https://snack.expo.io/SJD3c-swG
```
import React, { Component } from 'react';
import { View } from 'react-native';
import { GiftedChat } from 'react-native-gifted-chat' // 0.3.0

export default class App extends Component {
  render() {
    const userId= '123';
    const messages = [{
      _id: '1',
      createdAt: 1519225196922,
      text: 'Hello',
      user: {
        _id: userId,
        name: "Name",
      },
    }]
    return (
      <View style={{flex:1}}>
        <GiftedChat
          messages={messages}
          user={{ _id: userId }}
          dateFormat={'dddd DD. MMMM'}
          locale={'nb'}
        />
      </View>
    );
  }
}
```

#### Expected Results

That the time separator should be in norwegian (according to locale prop)

#### Additional Information

* Nodejs version: 7.5.0
* React version: 16.2.0
* React Native version: 0.52
* react-native-gifted-chat version: 0.3.0
* Platform(s) (iOS, Android, or both?): At least on iOS (haven't tested android)

jaaberg:
If the depending project uses another version of moment, importing languages won't work because it will be created two different instances. This PR adds moment to peerDependencies to ensure that the same version om moment is being used. 

Resolves #762. 

edit: I started by just moving it to peerDependecies, but since this repo also needs moment, this broke a lot of tests. I figured I could have moment as both a dependency and a peerDependecy. I'm not 100% sure that this is correct, but that was my best guess.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,5 +59,8 @@
     "react-native-parsed-text": "^0.0.20",
     "shallowequal": "1.0.2",
     "uuid": "3.2.1"
+  },
+  "peerDependencies": {
+    "moment": "^2.19.0"
   }
 }


xcarpentier:
LGTM

PeterKottas:
@jaaberg Just make it a dev dependency, that will work as expected.

jaaberg:
@PeterKottas I also thought about that, but doesn't `devDependency+peerDependency` mean that the consumer _must_ install moment themselves, but if we use `dependency+peerDependency` we say that this will work even if you don't install moment yourself, but if you install moment, please use this version?

PeterKottas:
Well here's the thing. Normally (on web) libraries are bundled. That means that you have to use peer dep otherwise the package ends up  bundled with rest of the code. That is obviously totally undesirable. However with native, libraries for some reason shipped not only unbundled, but even uncompiled. However in my mind nothing changes. What you should do is (afk pseudolike code):
```
let moment = undefined;
moment = require('moment').default;
if(!moment) {
  throw new error('Please install moment library');
}
```
So yes, you allow (and force) people to install their own, but this is the only way to give them control over the library version. While it might seem a bit unfriendly towards user, I've used this approach in each of my libs and seen it done all over and it definitely works. 

Plus this way, you can even do fallbacks. So instead of error, log warning, and fallback to some custom, simplified formatting. (once again, very popular on web because people might want to avoid pulling down a big lib like moment just to format one datetime somewhere)

jaaberg:
Wow, time flew by, thanks a lot for the explanation @PeterKottas, this was unknown territory for me. I will let this be up to the maintainers to decide what to do. I don't have time to push an alternative PR right now with the solution you suggest, so if anyone else would like to do it, feel free :)

Frexuz:
We really need this PR :) Seems many people (including me) are still having locale issues.
https://github.com/FaridSafi/react-native-gifted-chat/issues?utf8=%E2%9C%93&q=in%3Atitle+locale

n-sviridenko:
Is there any updates on it? Is this possible to make moment optional?

tiagoalves:
Would be great to have this merged.

In the meantime, if you use yarn, you can use its selective dependency resolutions feature. You can add the following to your `package.json` so that all `moment` dependencies resolve to the same version:

```
  "resolutions": {
    "moment": "2.22.2"
  }
```

sibelius:
how can we accept any moment version?

PR #776, opened 2018-03-01
Replace hardcoded width set to full screen width

attitude:
`InputToolbar` has hardcoded width set to full screen width.

That fails when `GiftedChat` is not used full screen, part of the input is hidden:

simulator screen shot - ipad air 2 - 2018-03-01 at 14 28 06

After applying the fix, input inherits container width:

simulator screen shot - ipad air 2 - 2018-03-01 at 14 26 39

### DIFF
--- a/src/InputToolbar.js
+++ b/src/InputToolbar.js
@@ -102,7 +102,8 @@ const styles = StyleSheet.create({
     borderTopColor: Color.defaultColor,
     backgroundColor: Color.white,
     bottom: 0,
-    width: Dimensions.get('window').width,
+    left: 0,
+    right: 0,
   },
   primary: {
     flexDirection: 'row',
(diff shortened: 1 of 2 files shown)


gianpaj:
@sujameslin can you check this approach?

is this better than what you proposed in https://github.com/FaridSafi/react-native-gifted-chat/pull/764?

sujameslin:
@gianpaj I can't judge that its better or not. 
Because finally what we want is full screened chat container where we don't see "green" area anymore.

attitude:
This should cover both full screen as well any partial/container approach too.

Motivation was to be able to adapt to a tablet layout when chat fills just ⅔ of the screen.

sujameslin:
@gianpaj I agree that this is better approach than #764. I'm closing my PR since its good. 
@attitude Yes, you're right. This addresses #761. Also would you please update jest snap so it can pass test in CircleCI?

sujameslin:
@FaridSafi This looks good to me and fixed my problem. 
Can we merge this PR?

PR #786, opened 2018-03-09
Fixed TS typings

ascariandrea:
Hi guys, I reworked what I did cause some of the typings where wrong and other missing.

To test this I did:
- `yarn pack` to create the tarball
- `yarn add path/to/tarball/react-native-gifted-chat` in my current project

All problem solved!

Let me know if I need to test it more


xcarpentier:
LGTM

PR #790, opened 2018-03-12
Fix a bug when custom actions is pressed

colloquet:
Because line 24 and line 26 are not using arrow function, `this` is not referring to the component itself. Therefore it throws `undefined is not an object` when trying to evaluate `this.props.options`.

Changing back to arrow function fixed the issue.

### DIFF
--- a/src/Actions.js
+++ b/src/Actions.js
@@ -21,9 +21,9 @@ export default class Actions extends React.Component {
         cancelButtonIndex,
         tintColor: this.props.optionTintColor,
       },
-      function handle(buttonIndex) {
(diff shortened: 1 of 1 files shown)


colloquet:
Hi I am unfamiliar with diff coverage concept, not sure why the codecov test is failing. I tried running the tests locally through `yarn run test` and it seems to have passed all 18 tests. Am I missing something in the pull request? 

Any help is greatly appreciated. 🙇‍♂️

xcarpentier:
LGTM

colloquet:
I am closing this PR because It is fixed in the latest release.

PR #791, opened 2018-03-13
Update Actions.js

dioptre:
Fixed popup actions in example.


xcarpentier:
@dioptre thanks
Can you fix lint issues?

Issue #797, opened 2018-03-17
Change color of Time text


pentarex:
#### Issue Description

I am trying to change the text color of the Time component, but sadly I cannot.
Before it was working when I set the textStyle of the Bubble... (in v0.3.0)
I saw that there was a change on 18 of December 2017 [5b67203570a2bcadf9475eec3c14149c4e404eac#diff-fde1e44c890e5f6e50b1f68f49f1034c](https://github.com/FaridSafi/react-native-gifted-chat/commit/5b67203570a2bcadf9475eec3c14149c4e404eac#diff-fde1e44c890e5f6e50b1f68f49f1034c)
and the this.props.textStyle was removed
What am I missing?

#### Steps to Reproduce / Code Snippets

```javascript
    /** */
    render() {
        return (
            <GiftedChat
                loadEarlier={this.state.loadEarlier}
                onLoadEarlier={this.onLoadEarlier.bind(this)}
                isLoadingEarlier={this.state.isLoadingEarlier}
                messages={this.state.messages}
                onSend={messages => this.onSend(messages)}
                user={this.getUserInfo()}
                renderFooter={this.renderFooter.bind(this)}
                onInputTextChanged={this.onInputTextChanged.bind(this)}
                renderBubble={this.renderBubble.bind(this)}
                renderAvatar={this.renderAvatar.bind(this)}
                renderTime={this.renderTime.bind(this)}
                showUserAvatar={true}
                showAvatarForEveryMessage={true}
                chatId={this.chatId}
                minInputToolbarHeight={50}
                bottomOffset={50}
            />
        );
    }

/** render the chat bubble */
    renderBubble(props) {
        return (
            <Bubble
                {...props}
                wrapperStyle={{
                    left: {
                        backgroundColor: Colors.chatReceiver,
                    },
                    right: {
                        backgroundColor: Colors.chatSender,
                    }
                }}
                textStyle={{
                    right: {
                        color: Colors.snow,
                        fontFamily: 'Montserrat-Light',
                        fontSize: 14
                    },
                    left: {
                        color: Colors.snow,
                        fontFamily: 'Montserrat-Light',
                        fontSize: 14
                    }
                }}
            />
        );
    }

/** render the time labels in the bubble */
    renderTime() {
        return (
            <Time
                textStyle={{
                    right: {
                        color: Colors.snow,
                        fontFamily: 'Montserrat-Light',
                        fontSize: 14
                    },
                    left: {
[...]

PR #798, opened 2018-03-17
fixing Time textStyle issue #797

pentarex:
The property textStyle was missing as a parameter in the Time function


pentarex:
thank you :)

PR #810, opened 2018-03-27
Fixed Day headers when inverted prop is false

### LINKED ISSUE #789: inverted={false} messes up the chat, day order.
I need my chats to start from the top when first rendered, so I would like to use the newly added inverted prop.

So I used it as 
inverted = {false}
and it seems my chats are seen from the top to bottom without inversion. 

But, the day indicator text for the chat messages is jumbled up somehow.
I think the changes needed for correct ordering wasn't accounted for when inverted prop is set to false.

img_2156

The behavior looks like the above picture.

Day "Mar 9, 2018" is supposed to be placed on top of the Obama picture. 
Day "Mar 11, 2018" is however in the correct position. 

It seems that if there are multiple chat messages for the same day, the day indicator text is rendered 
just above the last chat message of that given day. 
This obviously is a legacy from before when the inverted prop was not supported, can anyone fix this?

* react-native-gifted-chat version: 0.4.1

apparition47:
Resolves #789 due to a bug in PR #658.

### DIFF
--- a/src/Day.js
+++ b/src/Day.js
@@ -57,7 +57,6 @@ Day.defaultProps = {
   },
   previousMessage: {},
   nextMessage: {},
-  inverted: true,
   containerStyle: {},
   wrapperStyle: {},
   textStyle: {},
--- a/src/Day.js
+++ b/src/Day.js
@@ -56,6 +56,8 @@ Day.defaultProps = {
     createdAt: null,
   },
   previousMessage: {},
+  nextMessage: {},
+  inverted: true,
   containerStyle: {},
   wrapperStyle: {},
   textStyle: {},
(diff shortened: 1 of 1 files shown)


xcarpentier:
LGTM

cc @brunocascio ?

brunocascio:
LGTM

brzonsea:
Hi, I was the one who wrote the #789 issue. I had no problem with this PR fix until recently when I noticed that this fix has problems with the `renderAvatarOnTop` prop in the `Avatar` component.

I fixed the #789 issue differently, by changing the assignment of `previousMessage` and `nextMessage` inside the `prepareMessages` method of `MessageContainer` component. 

Is it ok if I submit a pull request with my fix?

PR #819, opened 2018-04-02
fix on action press

haqiqiw:
when i try example code, when i select one of the action, it run all the action.

so if i have 2 action (a & b) to show alert, when i select action a, it show 2 alert from action a and b.

thanks


rzdev:
Thank you for the fix, it's working.

PR #856, opened 2018-05-11
Dev - Keyboard: create props for handling keyboard animation

anhtukhtn:


anhtukhtn:
create prop for handling keyboard show/hide animation or not

PR #859, opened 2018-05-16
Fixed MessageContainer to account for reverse order of messages when inverted prop is false

### LINKED ISSUE #858: Wrong assignment of  previousMessage and nextMessage when inverted prop is false
I created an issue about a buggy Day headers when `inverted` prop is set to false back in #789.

This was fixed by #810 , but it seems the fix has some other issues.

For example, when I set `renderAvatarOnTop` for the `Avatar` component, the avatar is instead rendered at the botttom. 

This is because the current code does not switch the assignment of `previousMessage` and `nextMessage` when it first assigns them in `MessageContainer`.

I think the current assignment of  `previousMessage` and `nextMessage` has to be reversed when the `inverted` prop is set to false. 

I reversed these assignments in my own project, and it rendered the Day headers and `renderAvatarOnTop`  correctly.

brzonsea:
I think this resolves #858 .  Due to bugs in PR #658 and #810 .

### DIFF
--- a/src/Day.js
+++ b/src/Day.js
@@ -14,7 +14,7 @@ export default function Day(
   { dateFormat, currentMessage, previousMessage, nextMessage, containerStyle, wrapperStyle, textStyle, inverted },
   context,
 ) {
-  if (!isSameDay(currentMessage, inverted ? previousMessage : nextMessage)) {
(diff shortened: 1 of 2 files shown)


reimertz:
I can confirm this fixed same issues I was seeing. Great work @brzonsea!

PR #861, opened 2018-05-17
Use IMessage type instead of `any` to define the type of message property

arayaryoma:
Hello. 
I'm using react-native-gifted-chat with TypeScript. I want `messages` property of GiftedChat component to have a type.
 I considered to write new Interface but found `IMessage` type in the type definition file. It looks good to use as type of `messages` property.
Thanks.


xcarpentier:
#1190

PR #879, opened 2018-06-01
Update using multi language

imcodervn:
multi language and fix lag when type text input message


src/Composer.js
```diff
@@ -43,7 +43,7 @@ export default class Composer extends React.Component {
         onChangeText={(text) => this.onChangeText(text)}
         style={[styles.textInput, this.props.textInputStyle, { height: this.props.composerHeight }]}
         autoFocus={this.props.textInputAutoFocus}
-        value={this.props.text}
+        //value={this.props.text}
```
xcarpentier: Why?

PR #898, opened 2018-06-08
Fix MessageContainer keyboard avoiding when messages array is empty

doomsower:
I guess this is a regression which appeared after recent migration to `FlatList`.
When I open a chat without messages and want to type in the first message, keyboard appears and blocks `InputToolbar`. This happens because `MessageContainer` doesn't know what to do with props like `onKeyboardWillShow` in case when messages array is empty. When array is not empty, these props are passed down to list view. 
I added direct keyboard listeners in case when messages array is empty.

### DIFF
--- a/example-slack-message/src/App.js
+++ b/example-slack-message/src/App.js
@@ -1,12 +1,38 @@
-import React, { Component } from 'react';
-import { View, Platform } from 'react-native';
+import React from 'react';
+import { Platform } from 'react-native';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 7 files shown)


qwert666:
Still the case? seems like an important PR

kesha-antonov:
Experiencing same https://github.com/FaridSafi/react-native-gifted-chat/issues/923

doomsower:
Oh, I totally forgot about this PR. Now linter is happy, and it is probably ready get merged.
@xcarpentier can you review this?

PR #705, opened 2018-01-15
Modified for flatlist (from #629)

xcarpentier:
From https://github.com/FaridSafi/react-native-gifted-chat/pull/629

PR #914, opened 2018-06-28
fix: enable renderFooter for FlatList MessageContainer

glebez:
This PR fixes renderFooter gifted chat prop not working with the new Flat List implementation of Message Container (#705 ).

Flat List, compared to List View used before, does not recognise `renderFooter` and `renderHeader` props, so, although they were passed, they were ignored.

This PR removes above mentioned props and utilises `ListHeaderComponent` prop which is an alternative of `renderHeader` for Flat List.

### DIFF
--- a/src/MessageContainer.js
+++ b/src/MessageContainer.js
@@ -104,10 +104,9 @@ export default class MessageContainer extends React.PureComponent {
           style={styles.listStyle}
           contentContainerStyle={styles.contentContainerStyle}
           renderItem={this.renderRow}
-          renderHeader={this.renderFooter}
-          renderFooter={this.renderLoadEarlier}
           {...this.props.invertibleScrollViewProps}
           ListFooterComponent={this.renderHeaderWrapper}
+          ListHeaderComponent={this.renderFooter}
           {...this.props.listViewProps}
         />
       </View>


glebez:
@xcarpentier Thanks a lot for the quick review and merge! I really appreciate what you do here as a maintainer, thanks for all your effort!

src/MessageContainer.js
```diff
@@ -104,10 +104,9 @@ export default class MessageContainer extends React.PureComponent {
           style={styles.listStyle}
           contentContainerStyle={styles.contentContainerStyle}
           renderItem={this.renderRow}
-          renderHeader={this.renderFooter}
-          renderFooter={this.renderLoadEarlier}
           {...this.props.invertibleScrollViewProps}
           ListFooterComponent={this.renderHeaderWrapper}
+          ListHeaderComponent={this.renderFooter}
```
nandiniparimi1107: Am i mistaken or are the properties swapped? Shouldn't  ListFooterComponent={this.renderFooter}?

glebez: the list is inverted, so header becomes footer and vice versa. So they are swapped on purpose

PR #929, opened 2018-07-17
Show the username on the message

arilton:
If the prop renderUsernameOnMessage is setted to true, the message user name will be showed on the bubble. Else, nothing happens.


AlanQuadros:
Waiting for this PR...

dengue8830:
why this is not merged? has passed a month

dengue8830:
it seems the authors don't want this feature, [here](https://github.com/FaridSafi/react-native-gifted-chat/pull/448#issuecomment-423768022) is my faster solution

Issue #994, opened 2018-10-12
ISystemMessage is defined twice in index.d.ts


raphaelpra:
#### Issue Description

ISystemMessage is defined twice: 
- as a chat message data 
- as a component

Since only the last definition stands, whenever we need to defined a system chat data in our component (for instance in our state), it don't point to the correct element

#### Steps to Reproduce / Code Snippets

The following code raise an error :
```
import { ISystemMessage } from 'react-native-gifted-chat'

const systemMessage: ISystemMessage = { _id: "2",
            text: 'This is a system message',
            createdAt: new Date(),
            system: true,
          }
```

#### Expected Results

Not raising the following error ->
```
Type '{ _id: string; text: string; createdAt: Date; system: true; }' is not assignable to type 'ISystemMessage'.
  Property 'setState' is missing in type '{ _id: string; text: string; createdAt: Date; system: true; }'.
```

#### Additional Information

with typescript
* Nodejs version: not relevant
* React version: not relevant
* React Native version: not relevant
* react-native-gifted-chat version: 0.4.4
* Platform(s) (iOS, Android, or both?): both

raphaelpra:
I have suggested a renaming in a PR

xcarpentier:
#1190

PR #995, opened 2018-10-12
Change component interface name ISystemChat to SystemChat

raphaelpra:
ISystemChat was defined twice leading to issues


raphaelpra:
Fixing #994

PR #1000, opened 2018-10-19
fix android  keyboardWillShow not effect in InputToolbar

iamsee:
android's InputToolbar can't show when keyboard show


androsanta:
I'm experiencing this behaviour too. Replacing keyboardWillShow event width keyboardDidShow solve this, as pointed out here [https://facebook.github.io/react-native/docs/keyboard#addlistener](url).

tapz:
Any chance this could be merged soon?

PR #1010, opened 2018-11-06
add send video support 

maherzaidoune:


sibelius:
I’m not sure if this is the right direction t go

I think we should just provide a renderMessage and let user decide what to render

It is impossible to cover all types of messages

ahartzog:
Just like image is supported though, video is a very common use case.

Moneemsaadaoui:
Works like a charm and honestly video support makes life a tad bit easier 🐜🐜

sibelius:
CI is failing, can you fix it?

maherzaidoune:
@sibelius this is a eslint no-use-before-define failure because of the styles variable , while it's a common pattern to put the styles at the bottom of the component file , any suggestion ?

sibelius:
can we disable this lint rule in this specific case?

maherzaidoune:
done

PR #1047, opened 2018-12-20
isLoadingEarlier prop bug fix

jsartisan:
loadEarlier component was not re-rendering when the props were changing because FlatList only re-render when the props changes but in the current code, the props were same, so had to pass the isLoadingEarlier={this.props.isLoading} to FlatList of Message Container.

### DIFF
--- a/src/MessageContainer.js
+++ b/src/MessageContainer.js
@@ -136,6 +136,7 @@ export default class MessageContainer extends React.PureComponent {
           ListFooterComponent={this.renderHeaderWrapper}
           ListHeaderComponent={this.renderFooter}
           {...this.props.listViewProps}
+          isLoadingEarlier={this.props.isLoadingEarlier}
         />
       </View>
     );


src/MessageContainer.js
```diff
@@ -136,6 +136,7 @@ export default class MessageContainer extends React.PureComponent {
           ListFooterComponent={this.renderHeaderWrapper}
           ListHeaderComponent={this.renderFooter}
           {...this.props.listViewProps}
+          isLoadingEarlier={this.props.isLoadingEarlier}
```
maherzaidoune: can you add isLoadingEarlier to propTypes to avoid  ci/circleci problem ??

EgidioCaprino: [FlatList](https://facebook.github.io/react-native/docs/flatlist#props) does not seem to have an `isLoadingEarlier` prop

EgidioCaprino: `FlatList` takes an `extraData` prop for forcing re-rendering. Check [this](https://facebook.github.io/react-native/docs/flatlist#extradata)

jsartisan: Yes that would work i think.

Hirbod:
Nice.. that Stalebot starts to close real issues.. is this repo still maintained?

PR #1072, opened 2019-01-07
Fixes failed prop type warning for resizeMode

alvincrespo:
## Changes

- Removes resizeMode from `videoStyle` props as this causes `Warning: Failed prop type: Invalid props.style key resizeMode`


alvincrespo:
w00t! Thanks @xcarpentier

Issue #1048, opened 2018-12-20
Too large list of message fail to display


xcarpentier:
#### Issue Description

Too large list of message fail to display

#### Steps to Reproduce / Code Snippets

try a big list of message

#### Expected Results

Seeing something

#### Additional Information

* Nodejs version: 10
* React version: 16
* React Native version: 0.57
* react-native-gifted-chat version: 0.5.0
* Platform(s) (iOS, Android, or both?): both

PR #1076, opened 2019-01-09
Improve perf

xcarpentier:
https://github.com/FaridSafi/react-native-gifted-chat/issues/1048


xcarpentier:
@sibelius because that cause issue on large list of message (+200). Too much memory usage due to the deep shallow equal (blank screen, crash).

xcarpentier:
@sibelius do you have any other suggestions ?
BTW we passed thought all component huge amount of props and shallowequal did the work and consume too much memories that our thread not display any UI. (ie. https://github.com/FaridSafi/react-native-gifted-chat/pull/1076)

sibelius:
Good enough for me

brunocascio:
Is expo snack working? Code looks good for me

ChrisEdson:
This PR is super buggy, for a bunch of reasons.

It means that if you change any property in a message (like pending etc), the FlatList will not re-render because of the `shouldComponentUpdate` in MessageContainer.js. As the actual length of the messages array isn't changing, then nothing will update.

There's also bigger isssues in the PR - in that `shouldComponentUpdate` function, `next.extraData` and `current.extraData` is referenced. `next` is `this.props.messages` - so you're referencing the `extraData` property on a messages array. The same for `loadEarlier`.

I highly recommend using flow for things like this, it would have picked this up immediately - rather than merging something in that will break everyone's code.

xcarpentier:
@ChrisEdson please review: https://github.com/FaridSafi/react-native-gifted-chat/pull/1101

ChrisEdson:
Ah sorry my comment sounded really mean! 

I just mean adding Flow (or TypeScript). Then it would flag that you were trying to access `extraData` on an Array type (`messages`). TypeScript would do the same, but is a bit more work to add. Flow is pretty quick!

xcarpentier:
@ChrisEdson TypeScript is just better. I really don't want to introduce Flow.

So migrate to TypeScript?!? 
Challenge accepted!

Just need to know if eough contributor will continue or not?

**So please vote here with a 👍 if you see it ;)**

xcarpentier:
#1190

PR #1078, opened 2019-01-11
 Updating the code to fix MinComposerHeight is not respecting props in some instances.

nandiniparimi1107:
## Issue
minComposerHeight is hardcoded to use constant instead of props in few instances. Due to this minimumToolbarHeight will give incorrect result thus making UI look inconsistent.

## Solution
updating the code to use the props in all instances.

## Testing
if we add minComposerHeight prop to the giftedChatComponent, there is a bottomOffset on keyboard open as the minInputToolbarHeight is returning incorrect value based on the hardcoded minComposerHeight.

### DIFF
--- a/src/GiftedChat.js
+++ b/src/GiftedChat.js
@@ -55,7 +55,7 @@ class GiftedChat extends React.Component {
 
     this.state = {
       isInitialized: false, // initialization will calculate maxHeight before rendering the chat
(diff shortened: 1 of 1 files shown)


nandiniparimi1107:
@xcarpentier Please take a look at this.

xcarpentier:
Ok I see.
I agree with your change. I will merge in master but have no time to test.
Please give your test feedback.

PR #1094, opened 2019-01-22
Fix extraData is always undefined

MarkRunWu:


xcarpentier:
Sorry this component changed.

PR #1101, opened 2019-01-24
fix MessageContainer-Pure

xcarpentier:
To fix this: https://github.com/FaridSafi/react-native-gifted-chat/pull/1076#issuecomment-457158190
cc @ChrisEdson


ChrisEdson:
Looks good to me!

xcarpentier:
@ChrisEdson thanks for your feedback

PR #1109, opened 2019-01-28
Call scrollToBottomComponent function

zmefz:
`scrollToBottomComponent` parameter is mentioned as a function at PropTypes but it is not called.
This change will just call this function


zmefz:
`scrollToBottomComponent` should be a function (according to PropTypes). It means that there are two options:
1. Fix [propTypes](https://github.com/zmefz/react-native-gifted-chat/blob/15256a0d795ef1e7540246c5a5ed53c6e7f4a659/src/MessageContainer.js#L244) and change the type of the property to react node
2. Leave this property as a function and call it in place (what I did at the PR). Also in the case it will be better to rename it to `renderScrollToBottomComponent` or `renderScrollToBottom`

PR #1123, opened 2019-02-05
Fix expo import

mattvv:
Pulling from latest I had an issue with my project importing expo, even though isExpo() is false. It turns out that the require is pulled out in the static analysis and executed first. So I just added some fluff to get past that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,7 @@ e.g. System Message
 * **`scrollToBottom`** _(Bool)_ - Enables the scrollToBottom Component (Default is false)    
 * **`scrollToBottomComponent`** _(Function)_ - Custom Scroll To Bottom Component container  
 * **`scrollToBottomOffset`** _(Integer)_ - Custom Height Offset upon which to begin showing Scroll To Bottom Component (Default is 200)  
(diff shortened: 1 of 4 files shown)


xcarpentier:
I think the expo thing is not a good path by the way. I will remove it as soon as possible because we still have the possibility to use renderVideo, and I guess it should work also.

mcmar:
@mattvv @xcarpentier What is the last good version of this library? I'm currently unable to compile with this lib unless I'm using expo. Just need to figure out the last working version so I can downgrade to that

mattvv:
I have to be honest. I had the same issue @mcmar which is why I fixed this, as I was implementing an unrelated feature #1124

xcarpentier:
Expo thing is not released at all (left = `0.7.2`; right = master): https://github.com/FaridSafi/react-native-gifted-chat/compare/9db9d64...master

@mcmar you have to use `0.7.2`. Do you have any trouble with `0.7.2` version?

As I said I will remove Expo import because it's not required, we can use `renderMessageVideo` instead.

Can you both please copy past here your error on your build?

mattvv:
@xcarpentier it just fails trying to import expo, even though isExpo() is false, the require import is still pulled out of the block and attempted to be imported

xcarpentier:
@mattvv ok, I will remove it now :)

PR #1153, opened 2019-03-01
[Example][Expo] Bumped Expo example version to 32

EvanBacon:
# How

* Changed: `react-native-maps` -> expo import of `react-native-maps` (web support)
* Changed: `app.json` -> `Constants` (bare workflow support)
* Changed: Bump Expo 25 -> 32
* Changed: Bump React 16.2.0 -> 16.5.0

# Test Plan

Ran this locally with the latest Expo client. Can confirm all functionality is still working.

### DIFF
--- a/example-expo/CustomView.js
+++ b/example-expo/CustomView.js
@@ -1,7 +1,7 @@
 /* eslint-disable */
 import React from 'react';
 import { View, StyleSheet } from 'react-native';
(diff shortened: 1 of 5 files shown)


EvanBacon:
@xcarpentier you may need to update the snack as well, but that should be very straight forward. Just toggling the SDK version should be enough ❤️ https://snack.expo.io/@xcarpentier/gifted-chat

Issue #938, opened 2018-07-25
Need to auto-scroll to a particular message of a chat conversation


Thyix:
Good afternoon,

I have a chat messages thread and I want to scroll to a particular message by it's Id. I have searched and haven't found anything so far.

e.g. : I have a 'message search' scene in my application. When you enter something in the search bar, every messages from every contacts are fetched and I display only some that corresponds with the entry.  

When I click on a particular message of a particular contact, I navigate to this contact chat conversation. 

The only thing I need is scrolling to that particular message. I know every message have an Id, that's why I wanted to know if there's a method to move directly to a specific message by providing the chat Id.

Hope you can help me with this 👍

#### Additional Information

* Nodejs version: v8.9.4
* React version: 16.2.0
* React Native version: 0.55.4
* react-native-gifted-chat version: 0.4.3
* Platform(s) (IOS&Android): (IOS v11.2, Android v6.0)

leron8:
I will also be looking forward to having this feature. I see that gifted chat uses RN's FlatList (https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/MessageContainer.js) so this could be used https://facebook.github.io/react-native/docs/virtualizedlist.html#scrolltooffset Which I see they have that function in Gifted chat also:

scrollTo(options) {
    if (this.flatListRef) {
      this.flatListRef.scrollToOffset(options);
    }
  }

I haven't been able to dig in more to see how to implement it though. I hope it helps in some way!

alexvaltin97:
Also looking for it

flashimxd:
I'm looking for something similar...
Let me know if you guys figure out how to do that.

maulikdhameliya:
@leron8 : created this scrolling functionality ?

leron8:
Hello @maulikdhameliya I haven't implemented it. Been busy. I believe @Thyix might have worked on a PR, but not sure though.

Thyix:
Didn't find anything yet, sadly :(

testshallpasswork:
Technically, as alluded gifted chat uses Flatlist internally and it could be that you can use [scrollToIndex](https://facebook.github.io/react-native/docs/flatlist#scrolltoindex) but as documentation states it depends on [getItemLayout](https://facebook.github.io/react-native/docs/flatlist#getitemlayout) and that depends on giving Flatlist heights to the items upfront. 
@leron8 Your PR would be able to calculate variable item heights?

[...]

PR #1157, opened 2019-03-05
Need to auto-scroll to a particular message of a chat conversation

parthya:
re: https://github.com/FaridSafi/react-native-gifted-chat/issues/938

- Scroll to particular message using Flatlist's scrollToIndex (https://facebook.github.io/react-native/docs/flatlist#scrolltoindex). 

- Get the index of the message you want to scroll to in messageProps as `listIndex`. 

- Add ref to `<GiftedChat ref={(component) => (this._giftedChatRef = component)} />` and call `scrollToIndex` onPress using ref. For eg, `this._giftedChatRef.scrollToIndex({index: 8, viewOffset: 0, viewPosition: 1})`

### DIFF
--- a/src/GiftedChat.js
+++ b/src/GiftedChat.js
@@ -301,6 +301,14 @@ class GiftedChat extends React.Component {
     this._messageContainerRef.scrollTo({ y: 0, animated });
   }
 
+  // Trigger onPress using ref to <GiftedChat />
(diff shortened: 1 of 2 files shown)


LuckyShine001:
I want a scenario that on particular condition the chat should not scroll to the bottom(latest message), unless user manually scroll it. 
can we achieve it with updated gifted-chat version?

PR #1171, opened 2019-03-25
Improving keyboard performance using KeyboardAvoidingView

nandiniparimi1107:
## Issue
Keyboard animation is not smooth as expected. there is a bit of glitch when the keyboard shows up
keyboard_performance_no_view

## Solution
Wrapping MessageContainer in KeyboardAvoidingView would solve this issue.
keyboard_performance

Note:  In order for this to work, we need to make isAnimated = false. Since keyboardAvoidingView takes care of animating the view and keyboard, we don't need any custom logic to animate the view. 

In near future, we can deprecate the isAnimated prop based on the #issues

## Testing
1. Launch the app
2. click on the text input so that keyboard shows up.
3. Currently there is a glitch in the on show.

### DIFF
--- a/src/GiftedChat.js
+++ b/src/GiftedChat.js
@@ -8,7 +8,7 @@
 
 import PropTypes from 'prop-types';
 import React from 'react';
-import { Animated, Platform, StyleSheet, View } from 'react-native';
+import { Animated, KeyboardAvoidingView, Platform, StyleSheet, View } from 'react-native';
 
 import ActionSheet from '@expo/react-native-action-sheet';
 import moment from 'moment';
@@ -305,20 +305,21 @@ class GiftedChat extends React.Component {
   renderMessages() {
     const AnimatedView = this.props.isAnimated === true ? Animated.View : View;
     return (
-      <AnimatedView
-        style={{
-          height: this.state.messagesContainerHeight,
-        }}
-      >
-        <MessageContainer
-          {...this.props}
-          invertibleScrollViewProps={this.invertibleScrollViewProps}
-          messages={this.getMessages()}
-          ref={(component) => (this._messageContainerRef = component)}
-
-        />
-        {this.renderChatFooter()}
-      </AnimatedView>
+      <KeyboardAvoidingView enabled>
+        <AnimatedView
+          style={{
+            height: this.state.messagesContainerHeight,
+          }}
+        >
+          <MessageContainer
+            {...this.props}
+            invertibleScrollViewProps={this.invertibleScrollViewProps}
+            messages={this.getMessages()}
+            ref={(component) => (this._messageContainerRef = component)}
+          />
+          {this.renderChatFooter()}
+        </AnimatedView>
+      </KeyboardAvoidingView>
     );
   }
 


xcarpentier:
Sorry, there are too many conflicts because I migrate the language to TypeScript. Can you please resolve it?

llamaluvr:
Without more options for configuration, adding KeyboardAvoidingView to Gifted Chat could break many apps where GiftedChat is surrounded by navigation chrome. Typically, keyboardVerticalOffset needs to be set to the height of the header bar, or KeyboardAvoidingView needs to be entirely outside of the navigation container. See https://github.com/react-navigation/react-navigation/issues/3971. Keyboard avoidance can very on enough application-specific details that I think it's most flexible if it is left up to the consumer (other than what is already in there for iOS-specific content insets).

kesha-antonov:
Nice!

xcarpentier:
@nandiniparimi1107  What do you think about @llamaluvr said?

llamaluvr:
I can't say if I was reproducing the exact same scenario, but the other day I had tried adding KeyboardAvoidingView around basically the root component of my app, while removing my keyboard avoidance from all other views on iOS and Android (except for Gifted Chat, which has the iOS handling built in). I ran into situations where, the first time I would push a new view onto the stack with Gifted Chat and tap the text input, I would get double-accounting of the keyboard space, with the Gifted Chat text input flying a whole additional keyboard space above where it should be. It would only happen the first time I pushed the view, and only on my device (an iPhone SE). I couldn't reproduce it in Expo on a simulator, so it seemed like a timing issue with the animation, perhaps.

I think there's something to be said to switching to KeyboardAvoidingView, as, when I went into GiftedChat and commented out the animation based on keyboard height, I thought the pure KeyboardAvoidingView transition was a little smoother (hard to quantify/ explain- could be confirmation bias, but it seemed so!). It'd be great to have the option to disable GiftedChat's keyboard avoidance entirely. That way, on an app with pretty basic needs for keyboard avoidance, all it takes is adding KeyboardAvoidingView at the root and it's handled everywhere for both platforms. Once it's not optional (whether using the current custom solution or KeyboardAvoidingView), the keyboard avoidance has to be added per-screen so we explicitly don't add any double-avoidance on any chat screens.

walidvb:
Agreed with @llamaluvr , i have a `<KeyboardSpacer />` at the root of my (react-navigation) app, and it makes the input fill the entire chat/messages space.
I'll try and come up with a PR for disabling it, considering i just started looking at the code of GiftedChat yesterday.

walidvb:
indeed, as both `<KeyboardSpacer />` and `<GiftedChat />` handle it, i get an empty white space between the `InputToolbar` and the `MessageContainer`. (curiously, the spacing is only applied once a key was pressed. example [here](https://cl.ly/358fa489fbab))
Disabling the [callback](https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/GiftedChat.tsx#L541) fixes it.
Have made a PR regarding that particular issue [here](https://github.com/FaridSafi/react-native-gifted-chat/pull/1326)

shamilovtim:
Both this and #1326 are compatible together. They just need one more prop to provide an optional offset for people's navigation chrome. One thing I don't understand about this is I'm not sure why `AnimatedView` still exists since there's no animation anymore. Either way, if someone provides the keyboard avoiding view, an optional offset, and an optional way of disabling the keyboard avoiding view (letting a user manage that himself outside the chat) then this PR is ready to merge.

shamilovtim:
I should have a PR for this shortly @xcarpentier unless you have something better in mind

xcarpentier:
@shamilovtim ok ;)
I've got this issue in mind for a long time now but if you have a PR it will be awesome :D

shamilovtim:
I don't know if you've played with it yet but there are a lot of misconceptions. The way @nandiniparimi1107 added this it doesn't require handling keyboard offset the way @llamaluvr said and in theory it **shouldn't** require providing an option to disable the keyboardavoider the way that @walidvb requested. The reason being that in this PR the keyboardavoider isn't actually handling the keyboard but is instead simply smoothing the animation without the react-native Animation library. The positioning of the keyboard and chat still comes from the algorithm of gifted chat but it seems keyboardavoider is simply handling the animation in a better way. From playing with it my understanding is that nothing has changed here except for the fact that the animation looks superior.

@walidvb can you be clear, is your issue of #1326 only required when this PR is implemented or was it necessary even before from the original way that this plugin handled the form?

walidvb:
I'm really sorry, not sure what you mean.
That fix was needed on version `0.9.11` for sure, as far as i can tell the current patch(1171) was not part of that release. 
As mentioned on the other PR, i will test and revert within the next hour or two.

Thanks!

walidvb:
I can replicate on `0.11.4`, and the following patch fixes the issue:

```
diff --git a/node_modules/react-native-gifted-chat/.DS_Store b/node_modules/react-native-gifted-chat/.DS_Store
new file mode 100644
index 0000000..e69de29
diff --git a/node_modules/react-native-gifted-chat/lib/GiftedChat.js b/node_modules/react-native-gifted-chat/lib/GiftedChat.js
index a2029b3..fb0dbb7 100644
--- a/node_modules/react-native-gifted-chat/lib/GiftedChat.js
+++ b/node_modules/react-native-gifted-chat/lib/GiftedChat.js
@@ -51,6 +51,7 @@ class GiftedChat extends React.Component {
         };
         this.onKeyboardWillShow = (e) => {
             this.setIsTypingDisabled(true);
+            return
             this.setKeyboardHeight(e.endCoordinates ? e.endCoordinates.height : e.end.height);
             this.setBottomOffset(this.safeAreaIphoneX(this.props.bottomOffset));
             const newMessagesContainerHeight = this.getMessagesContainerHeightWithKeyboard();
```

PR #1190, opened 2019-04-05
Migrate GiftedChat to TypeScript

xcarpentier:
Migrate the project to TypeScript.

cc @slorber @EvanBacon  ;)


EvanBacon:
thanks for this

xcarpentier:
Hi @EvanBacon! What do you think if we change a bit the examples files structure and make a single app example by merging others?

And also do something like @slorber did every time when he makes an RN lib.
I mean an expo app at the root and an npm script before publish. It's really more easy for dev cycle of a lib! see it here:
- https://github.com/slorber/react-native-animation-hooks
- I copied his awesome idea too in my last packages: https://github.com/xcarpentier/react-native-suggester or https://github.com/xcarpentier/rn-verifcode

EvanBacon:
@xcarpentier yeah, I think one example would be best. It'd also be great to make things like react-native-video optional now that Expo (specifically expo-av) has web support.

xcarpentier:
OK, will do it in another PR.

@EvanBacon can you precise your suggestion about web support (ie. expo-av)?

src/MessageContainer.js
```diff
@@ -1,253 +0,0 @@
...
-
-  render() {
-    if (this.props.messages.length === 0) {
-      return <View style={styles.container} />;
-    }
-    return (
-      <View style={this.props.alignTop ? styles.containerAlignTop : styles.container}>
-        {this.state.showScrollBottom && this.props.scrollToBottom ? this.renderScrollToBottomWrapper() : null}
-        <FlatList
-          ref={(ref) => (this.flatListRef = ref)}
```
noelsp: It was good!

src/MessageContainer.tsx
```diff
@@ -0,0 +1,333 @@
...
+      <View
+        style={
+          this.props.alignTop ? styles.containerAlignTop : styles.container
+        }
+      >
+        {this.state.showScrollBottom && this.props.scrollToBottom
+          ? this.renderScrollToBottomWrapper()
+          : null}
+        <FlatList
+          ref={this.flatListRef}
```
noelsp: Here is an error. This means that when sending a new message, do not scroll down. I am right?

xcarpentier: I you have time make PR, it will be more explicit. Thanks

PR #1198, opened 2019-04-16
Add styles to render in user.avatar

jeronimo13:
I want to use `<CachedImage />` component instead of `<Image />` to use cached images on user device. I've noticed that `avatar` user props could be a function. 

Usage example:
```
   ...(m.author.profilePhotoUrl && {
      avatar: style => {
        return (
          <CachedImage
            source={{ uri: m.author.profilePhotoUrl }}
            style={style}
          />
        )
      }
    })
```


xcarpentier:
Thanks
I've copied it here https://github.com/FaridSafi/react-native-gifted-chat/commit/d85fe0457561f16920f919f53991a63d749fb9e4

PR #1208, opened 2019-04-28
Add alignTop to the GiftedChat prop typescript def

gsdatta:


xcarpentier:
Sorry this is obsolete

Issue #1192, opened 2019-04-06
How to use scrollToBottomComponent?


ckwee-alex:
#### Issue Description

I'm not sure how to use `scrollToBottomComponent` to customize the component with `scrollToBottom={true}`.

Got the following error when trying to use this two props together:
`ExceptionsManager.js:82 Warning: Functions are not valid as a React child. This may happen if you return a Component instead of <Component /> from render. Or maybe you meant to call this function rather than return it.`

#### Steps to Reproduce / Code Snippets
```
private _scrollToBottomComponent(props) {
  return (
    <View style={styles.scrollToBottomContainer}>
      <TouchableOpacity {...props}>
        <Text>Image here</Text>
     </TouchableOpacity>
   </View>
  );
}
```

#### Expected Results

To be able to center the component and change the current <Text>V</Text> to an image or possibly animation with either lottie-react-native or react-native-element. 

#### Additional Information

* Nodejs version: 11.13.0
* React version: 16.8.3
* React Native version: 0.59.3
* react-native-gifted-chat version: 0.7.2
* Platform(s) (iOS, Android, or both?): Both

wincjh:
+1

junminleeKS:
If you pass in the component instead of a function then it works although you get a warning from PropTypes.

ckwee-alex:
I noticed that as well. Is there any way to allow to change the content wrapped by `TouchableOpacity` which in this case is the `<Text>V</Text>`.

msqar:
Hey guys, i found there's no way to customize the container, I don't want it to be a circle-shaped with opacity. How can I edit the container? in the props it says i can pass the scrollToBottomStyle, but it's not using it anywhere within the code.

ismdcf:
Hi
Any update on this

PR #1209, opened 2019-04-29
Fix scroll to bottom usage

milanito:
# Description

As mentionned in #1192 , scrollToBottom usage does not work as expected

# Changes

- Used function call
- Used same container
- Allowed custom style


RETFU:
Thx @milanito it is the exact case we have at my job.

nicolas-adamini:
Same here !

milanito:
@xcarpentier any possibility you could check it ?

PR #1211, opened 2019-04-30
Bot message implementation (quick replies copy and inspiration)

xcarpentier:
## Description 
Start to provide a solution to add quick replies list component between `MessageContainer` and `Composer`.

Image-1 (1)

## Limitations
- In this first implementation, we just provide a single (radio) and multiple (checkbox) selections of option present to the user.

## Data definition and example
```ts

interface Reply {
  title: string,
  value: string
}

interface QuickReplies {
  type: 'radio' | 'checkbox',
  values: Reply[]
}
```
```js
{
  _id: 123,
  text: 'This is a quick reply. Do you love Gifted Chat?',
  createdAt: new Date(),
  quickReplies: {
    type: 'radio', // or 'checkbox',
    values: [
      {
        title: '😋 Yes',
        value: 'yes',
      },
      {
        title: '📷 Yes, let me show you with a picture!',
        value: 'yes_picture',
      },
      {
        title: '😞 Nope. What?',
        value: 'no',
      },
    ],
  },
}
```
### Pull-requests
- [WIP Quick Replies](https://github.com/FaridSafi/react-native-gifted-chat/pull/886)
- [Initial commit updating README.md to include new Quick Reply feature.](https://github.com/FaridSafi/react-native-gifted-chat/pull/881)
- [Add support for onPress on Bubble](https://github.com/FaridSafi/react-native-gifted-chat/pull/984)
### Issues
- [RFC: Quick Replies](https://github.com/FaridSafi/react-native-gifted-chat/issues/826)
- [Rendering Custom Responses (Like a Bot)](https://github.com/FaridSafi/react-native-gifted-chat/issues/367)

Closes #826

PR #1234, opened 2019-05-20
missing object user in message

jkmf:
Object user is required in order to render ticks and to fix showUserAvatar prop.


xcarpentier:
will be fixed on #1211

PR #1252, opened 2019-06-01
Changed day title format, Show actions after composer

azmatullahqau:


xcarpentier:
Sorry but I don't see any enhancement here.

PR #1270, opened 2019-06-13
Fixed QuickReply issue when values were identical

superphil0:
Changed keys for Quickreply react components from value to messageId

### DIFF
--- a/src/QuickReplies.tsx
+++ b/src/QuickReplies.tsx
@@ -152,7 +152,7 @@ export default class QuickReplies extends Component<
                 { borderColor: color },
                 selected && { backgroundColor: color },
               ]}
-              key={reply.value}
+              key={reply.messageId}
             >
               <Text
                 numberOfLines={2}


src/QuickReplies.tsx
```diff
@@ -152,7 +152,7 @@ export default class QuickReplies extends Component<
                 { borderColor: color },
                 selected && { backgroundColor: color },
               ]}
-              key={reply.value}
+              key={reply.messageId}
```
xcarpentier: That will cause the same error but for every reply because messageId is the same for every reply. I think we should keep value but maybe add index.
```tsx
<TouchableOpacity key={`${reply.value}-${indexFromMap}`}>
```

superphil0: Ah well there is a messageId prop for every reply, is set it explicitly in my test, would need to make messageId mandatory then not optional 

yes your solution is probably good as well, however i would not use the minus there since you also strings for the messageId (at least in the propTypes)

xcarpentier: So what you will do with this PR. Will you change the key or not?
TY

superphil0: I have to take a closer look next week, I saw that messageId gets overwritten internally.
So i don't understand currently why messageId can be set for replies at all.

also after I made changes as suggest in this PR when testing the react error with same keys disappeared, so for me the change works, but will have to write a clean test

xcarpentier: I thought we need to add `messageId` for a reply or replies is/are useful because you will need it on the handling the `onPress` event and pass the message id through the reply object.

Please refer in code what you mean by:
> overwritten internally

Thanks

xcarpentier:
Any news?

xcarpentier:
Will be fixed here: #1284

PR #1276, opened 2019-06-20
Typing improvements

hammadj:
Currently, all the components have generic props but don't actually make use of the generic, it always defaults to `IMessage`. This allows users to have a custom `Message` type (as long as it extends `IMessage`).

I also changed the utils to allow null/undefined values for diffMessage. The type of `props.previousMessage` is an optional, so we should allow passing in a potentially null/undefined value for diffMessage to make it easier/cleaner to use without having to do a check on `props.previousMessage` first.

Lastly, the type of style props should be wrapped in `StyleProp<>` in order to handle the various types of styles you can get from StyleSheet.create/inline styles, etc.

### DIFF
--- a/src/Actions.tsx
+++ b/src/Actions.tsx
@@ -6,6 +6,7 @@ import {
   TouchableOpacity,
   View,
   ViewPropTypes,
+  StyleProp,
   ViewStyle,
   TextStyle,
 } from 'react-native'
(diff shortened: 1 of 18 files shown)


hammadj:
@xcarpentier Could we bump the version and publish these changes to NPM?

xcarpentier:
Yes, I will try asap (maybe tomorrow)

hammadj:
> Yes, I will try asap (maybe tomorrow)

Thanks, appreciate it 👍

Issue #1246, opened 2019-05-27
How to send a view with Buttons on Gifted-Chat ? 


AtdheUruqi:
#### Issue Description

Hello guys,

how can i send a view with buttons something like this ? 
for example: when i click the hand on the input, it will send a view with buttons !

![GIThelp](https://user-images.githubusercontent.com/43423742/58422250-62a3e700-8092-11e9-95ab-05607ec35905.PNG)

xcarpentier:
You’ve to use `renderCustomView`

AtdheUruqi:
@xcarpentier now it render every-time i type something!
How can i manage that ?
I want when the user type and when it hit send now, to not render the CustomView that i made :/
![Screenshot_1559234946](https://user-images.githubusercontent.com/43423742/58649168-b5d49e80-830b-11e9-99b0-55a67dc7f8b1.png)

xcarpentier:
`renderCustomView(currentMessage)` you have to test your data in message when you want to display custom view from message meta.

AtdheUruqi:
@xcarpentier not sure i truly understand :/ , any example would be a great help .

evertonco:
It's very simple. Create a custom field on your message object, like "invitation". On rendercustomview method, check if currentmessage own invitation field, eg if(currentmessage.invitation). Otherwise, returno null

Issue #1201, opened 2019-04-23
position of "CustomView" component


3luyka:
#### Issue Description

I want to render `CustomView` below the text message. Could you please provide prop for that so I can choose position for `CustomView`?

#### Additional Information

* Nodejs version: 11.7
* React version: 16.8.3
* React Native version: 0.59.5
* react-native-gifted-chat version: 0.7.3
* Platform(s) (iOS, Android, or both?): both

evertonco:
renderCustomView allows you render whatever you want. So you can do something like: 
```
renderCustomView = ({ currentMessage, ...otherProps }) => {
  return (
    <View>
      <Text>{currentMessage.text}</Text>
      <YourCustomContentComponent />
    </View>
  )
}
```

PR #1298, opened 2019-07-05
Added isCustomViewBottom Prop

Lsleiman:
Seems like this feature request keeps coming back. Most frequent use case is having Accept/Reject buttons rendered underneath the message text.

Users can now determine if their custom view is rendered before or after bubble message, image or video.


Lsleiman:
See #1246 and #1201

Lsleiman:
I went ahead and changed the propType to Boolean with a default value of false. Let me know if i missed anything.
CustomView will display at the top unless users specifically set renderCustomViewBottom = true.

xcarpentier:
Sorry but I don't think this is a cristal clear name of props `renderCustomViewBottom`.
When `boolean` props I prefer I prefer something like  `isCustomViewBottom`.

Lsleiman:
mmm, I basically used _renderAvatarOnTop_, but will change and resubmit.

robertvorthman:
Did isCustomViewBottom get compiled into the lib .js files?   I can't get isCustomViewBottom to work and it does not appear appear in the Bubble.js file at node_modules/react-native-gifted-chat/lib/Bubble.js.

andersravn:
Was this ever included? Would be really nice to be able to add custom view such as buttons after the text.

PR #1319, opened 2019-07-16
Add containerStyle also to QuickReplies

superphil0:
Added containerStyle to Quickreplies in order to enable overwriting of hard style assumptions
(e.g. flexDirection row instead of column)
Real life use case, enable quick replies that are on top of each other and not next to each other

photo5868279302347403518

Please review if there is any other way of implementing this behaviour without overriding renderQuickReplies in Bubble or of the intention if this library is to provide this capability only with overriding


itsezc:
I agree with this also @superphil0 how did you achieve the result shown in the screenshot? could you post come code examples?

PR #1324, opened 2019-07-17
Features/1301 on long press avatar

walidvb:
Apologies for the previous PR that did not pass the tslint test.
Btw, not sure how i was able to even commit? Were the pre-commit hooks added just now?


xcarpentier:
I already include your PR for the next release and fixed the issue

walidvb:
Apologies, @xcarpentier, it [doesn't seem to me](https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/GiftedAvatar.tsx#L65) that this has been fixed in master? 
Please forgive me if I have missed smth.

I am currently unable to commit on another branch because of the hooks that have been added, and the fact that this is missing.

PR #1326, opened 2019-07-18
adds option to disable keyboard sizing handling

walidvb:
This is to fix fix some issues when keyboard is handled outside of the `GiftedChat` container, as discussed in https://github.com/FaridSafi/react-native-gifted-chat/pull/1171.
In my case, I use a `KeyboardSpacer` at the very root of my app, so I don't need this library to deal with that. 
A work-around would have been to wrap my root into a context, and disable the `KeyboardSpacer` on any view that presents a `GiftedChat`.

Hope you find this useful, and up to your coding standards 👍🏻

Any comments welcome, always.

---------------
_This also includes fix done in https://github.com/FaridSafi/react-native-gifted-chat/pull/1324_

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,6 +317,7 @@ interface QuickReplies {
 - **`listViewProps`** _(Object)_ - Extra props to be passed to the messages [`<ListView>`](https://facebook.github.io/react-native/docs/listview.html); some props can't be overridden, see the code in `MessageContainer.render()` for details
 - **`textInputProps`** _(Object)_ - Extra props to be passed to the [`<TextInput>`](https://facebook.github.io/react-native/docs/textinput.html)
 - **`keyboardShouldPersistTaps`** _(Enum)_ - Determines whether the keyboard should stay visible after a tap; see [`<ScrollView>`](https://facebook.github.io/react-native/docs/scrollview.html) docs
+* **`disableKeyboardHandling`** _(Bool)_ - Disable resizing of the messages container view when keyboard is shown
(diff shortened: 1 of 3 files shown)


shamilovtim:
any status on this?

walidvb:
Would also love an update on this, at least to know whether it's being considered at all ?

shamilovtim:
I'm working on a PR that incorporates this and the other related ticket

walidvb:
@shamilovtim any reason why not simply merge both? Or, are you planning on more/different changes?

shamilovtim:
The other branch was hundreds of commits behind plus had a merge conflict

shamilovtim:
@walidvb for clarity, did you need this when gifted chat uses absolute position and the Animation library or did you need this only after keyboardawareview got added in the other PR?

walidvb:
Hm my usecase is my root app has a `<KeyboardSpacer />`, therefore that spacing is handled by both. This is why i wanted to disable it in the chat. Alternative was to disable the `KeyboardSpacer` when a chat is present, but was more complicated.

shamilovtim:
Yeah I already understood that you have a keyboard handler near the root but what I don't understand was whether the current version of gifted chat causes you a problem, or whether only after the PR with the KAV gets merged would you have a problem?

walidvb:
@shamilovtim oh sorry,. I haven't tested around versions too much, but last i tried was `^0.9.11`. Will update now and revert 👌🏽

walidvb:
Thread moved to #1171

PR #1328, opened 2019-07-20
Feature/adding text position

jaygould:
Hey, I required a feature to see the position of the currently typed text inputted into the Composer's TextInput component. It is sent back to the Composer function via the onTextChanged prop so the position of the character is available on each key press. This is helpful in order to find out what the last typed character is in order to add user mentions (showing user list to select from in order to tag/mention as user). What do you think?

There was also a TS issue with the recent onLongPress prop for the avatar as the prop was not recognised by GiftedAvatar so I had a go at that.

Cheers!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,11512 @@
+{
+  "name": "react-native-gifted-chat",
+  "version": "0.9.11",
+  "lockfileVersion": 1,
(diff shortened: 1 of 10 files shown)


walidvb:
Indeed, part of dismissed  #1324, also fixed in #1326

Issue #1227, opened 2019-05-14
when set alignTop={true}, message got cut off by messageContainer when scrolling


crunchysoul:
#### Issue Description

when set alignTop={true}, messageContainer's height only set to match current messages, and when scrolling, message got cut off by messageContainer, for example, when only having less than full screen amount of messages.

#### Expected Results

messageContainer should behaviour like alignTop={false}, occupies full space when scrolling.

#### Additional Information

![recording](https://user-images.githubusercontent.com/20638592/57689328-aa824180-7682-11e9-88dc-858a758d5be9.gif)

* Nodejs version: v10.15.3
* React version: 16.6.3
* React Native version:  v0.58.4
* react-native-gifted-chat version: v0.8.2
* Platform(s) (iOS, Android, or both?): both
* TypeScript version: [FILL THIS OUT]

jsmile1994:
Same issue +1.

jan-wilhelm:
This still happens for me as well.

alorr10:
Anyone find a fix yet?

daviswhitehead:
I just created a PR to resolve this issue. Hopefully it gets merged soon.
https://github.com/FaridSafi/react-native-gifted-chat/pull/1330

alorr10:
@daviswhitehead your PR didnt work for me, I commented on the PR. Thank you for attempting tho!!

bonomoon:
I created a PR to resolve this issue. 
I think it's because of ios's overscroll feature. so I make block it when alignTop = {ture}.

jan-wilhelm:
This issue still exists and leads to bad UX. It should not be closed.
Can somebody please look into the mentioned PR? Thanks!

BrieucJ:
Worked for me:
alignTop={false} 
listViewProps={{contentContainerStyle: {flexGrow: 1, justifyContent: 'flex-end'}}}

lieberscott:
I've managed to figure this out if you still need to have `alignTop={ true }`. It requires a small tinker with the node-module itself.

Go to `node_modules > react-native-gifted-chat > lib > MessageContainer.js`.

Under `styles`, for `containerAlignTop`, add `flex: 1`. For `contentContainerStyle`, change` justifyContent: 'flex-start'` to `justifyContent: 'flex-end'`.

So it should look like this:

```
containerAlignTop: {
  flexDirection: 'row',
  alignItems: 'flex-start',
  flex: 1
},
[...]

PR #1330, opened 2019-07-22
quick fix to make messageContainer full screen when alignTop={true}

daviswhitehead:
resolves https://github.com/FaridSafi/react-native-gifted-chat/issues/1227


alorr10:
Same issue as @aguscha333 unfortunately. Thank you for trying tho

PR #1284, opened 2019-06-26
WEB Gifted Chat 

xcarpentier:
Using react-native-web for Gifted Chat.
Next version: [next](https://www.npmjs.com/package/react-native-gifted-chat/v/next)
@EvanBacon

PR #1359, opened 2019-08-15
[ActionSheet] Fix build with @expo/react-native-action-sheet@^3.0.3

kesha-antonov:
Hey!

Maybe it's better to use https://github.com/expo/react-native-action-sheet/blob/master/src/context.ts
IDK for now

At least this code will not break existing apps

Though we maybe need to wait while this PR gets merged https://github.com/expo/react-native-action-sheet/pull/127

Correct me please if I'm wrong

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "typescript": "3.4.5"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


xcarpentier:
I'm fixed it here #1284 
Will be merge today hopefully

Issue #680, opened 2017-12-26
Text box not visible when keyboard active using expo


hahahatrick:
#### Issue Description

Text bot is invisible when keyboard active in android with expo. I'm using```forceGetKeyboardHeight={true}``` and the message content moves correctly but there is an empty space instead of the textbox.
![screenshot_2017-12-26-18-51-19-590_host exp exponent](https://user-images.githubusercontent.com/33310702/34357541-0452d2d0-ea6e-11e7-8969-40171900f65d.png)

![screenshot_2017-12-26-18-50-57-697_host exp exponent](https://user-images.githubusercontent.com/33310702/34357545-0d6870fa-ea6e-11e7-8472-bef51173c3c7.png)

#### Steps to Reproduce / Code Snippets

```
<GiftedChat
        messages={this.state.messages}
        forceGetKeyboardHeight={true}
        onSend={(messages) => this.onSend(messages)}
        user={{
          _id: 1,
        }}
      />
```

#### Expected Results

The text box should be visible.

#### Additional Information

* Nodejs version: 8.9.3
* React version: 16.0.0
* React Native version: expo react native sdk 23.0.0
* react-native-gifted-chat version: 0.3.0
* expo version: 23.0.0
* Platform(s) (iOS, Android, or both?): Tested on android

xcarpentier:
And if you remove props forceGetKeyboardHeight?

hahahatrick:
@xcarpentier Then the screen does not get resized, the view remains the same and keyboard comes over it. I'm was able to fix it by removing the props forceGetKeyboardHeight and using KeyboardSpacer

AbhishekVG:
@hahahatrick even after putting KeyboardSpacer,The issue is not resolving.
Can you correct the below code if you have resolved the issue

```
<GiftedChat
        messages={this.state.messages}
        keyboardSpacer={true}
        onSend={(messages) => this.onSend(messages)}
        user={{
          _id: 1,
        }}
 />
```

bhavna18:
I had the same issue , but in my case textInputBar was pushing above the keyboard but its view was coming partially. 
` <KeyboardAvoidingView behavior={'padding'} style={{flex:1}} keyboardVerticalOffset={30}>
<ChatBot/>
</KeyboardAvoidingView>`
fixed my problem.
Here <ChatBot/> is my component where I have called GiftedChat .
Hope this helps.

wmonecke:
Same issue here also only on Android.

[...]

PR #1382, opened 2019-09-05
Update README on how to fix Android keyboard

chandlervdw:
I found this in another issue and I think it should be updated here. https://github.com/FaridSafi/react-native-gifted-chat/issues/680#issuecomment-377662861

### DIFF
--- a/README.md
+++ b/README.md
@@ -386,7 +386,7 @@ If you are using Create React Native App / Expo, no Android specific installatio
 ```
 <View style={{ flex: 1 }}>
    <GiftedChat />
-   {Platform.OS === 'android' ? <KeyboardAvoidingView behavior="padding" />}
+   {Platform.OS === 'android' && <KeyboardAvoidingView behavior="padding" />}
 </View>
 ```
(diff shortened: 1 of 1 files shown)


khanguslee:
I think you're missing the expression to return if your condition is false in your ternary operator.

Something like below would work instead?
```
   {Platform.OS === 'android' ? <KeyboardAvoidingView behavior="padding" /> : null}
```

README.md
```diff
@@ -382,16 +382,11 @@ If you are using Create React Native App / Expo, no Android specific installatio
...
 <View style={{ flex: 1 }}>
-   {
-      Platform.OS === 'android' ?
-         <KeyboardAvoidingView behavior="padding">
-            <GiftedChat />
-         </KeyboardAvoidingView> :
-         <GiftedChat />
-   }
+   <GiftedChat />
+   {Platform.OS === 'android' ? <KeyboardAvoidingView behavior="padding" />}
```
khanguslee: I think you're missing the expression to return if your condition is false in your ternary operator.

Something like below would work instead?

   {Platform.OS === 'android' ? <KeyboardAvoidingView behavior="padding" /> : null}

chandlervdw: d'oh! Thanks.

chandlervdw:
@khanguslee thanks for the feedback. Ternary fixed.

PR #1385, opened 2019-09-07
docs(README): Update KeyboardAvoidingView instructions

DAB0mB:
With the current instructions, the keyboard will hide the chat view.
With the new instructions, the keyboard will push the chat view as intended.


smxdevst:
On IOS view is hidden unless you start typing

Issue #1350, opened 2019-08-08
Locale prop not working


dreamchasersuon:
#### Issue Description

Init locale don't work as expected

#### Steps to Reproduce / Code Snippets

Look at this piece of code in src/GiftedChat:

![GiftedChat-initLocale](https://user-images.githubusercontent.com/34100648/62684003-23224100-b9d9-11e9-8d8e-cf1502b9f3bc.png)

This function break all things down... 
If you will debug this function like so: 

![GiftedChat-initLocale__debug](https://user-images.githubusercontent.com/34100648/62684140-77c5bc00-b9d9-11e9-8782-cebdc531da60.png)

You will get such a result: 

![GiftedChat-initLocale__debug-output](https://user-images.githubusercontent.com/34100648/62684284-ca9f7380-b9d9-11e9-93b7-5d6aed512ef7.png)

Every time check moment.locales().indexOf(this.props.locale) will give us 'en' and index '-1'. And no matter did we have locale props or not.
#### Expected Results

Expect to get the locale from props...

#### Additional Information

* Nodejs version: 11.15.0
* React version: 16.8.3
* React Native version: https://github.com/expo/react-native/archive/sdk-33.0.0.tar.gz
* react-native-gifted-chat version: ^0.5.0
* Platform(s) (iOS, Android, or both?): both
* TypeScript version: -----

PR #1391, opened 2019-09-11
locale prop fix

eminsr:
At the moment `moment.locales().indexOf(locale)` always returns `-1` for any locale other than `en`. Locale files needs to be imported first. `moment/min/locales` contains all the locale files so with this PR changing the locale prop will take affect.

#1350

### DIFF
--- a/src/GiftedChat.tsx
+++ b/src/GiftedChat.tsx
@@ -12,6 +12,7 @@ import {
 
 import ActionSheet from '@expo/react-native-action-sheet'
 import moment from 'moment'
+import 'moment/min/locales';
 import uuid from 'uuid'
 import { isIphoneX } from 'react-native-iphone-x-helper'
 


xcarpentier:
Hi, Thanks,
I'm not sure that is performant to embed every locale.
You may try to define `moment` yourself to find a version that fixes your issue.

eminsr:
Hi @xcarpentier,
Using same version of the `moment` not working anymore (for me at least). Since we can't `require` the `locale` dynamically maybe we could use switch case or something to `require` the `locale` in the `locale` prop before `moment.locales().indexOf(locale)`.

xcarpentier:
But you still able to change `moment` version since it's declared in peerDependencies.
https://github.com/FaridSafi/react-native-gifted-chat/blob/28114c3def965d2f68df6d29aed74eed6e5967c1/package.json#L110-L112

eminsr:
I'll try v10.1 and get back to you on that in couple days.

eminsr:
I tested v0.11.0 requiring locale works now.

PR #1436, opened 2019-10-02
feat: shouldUpdateChat for GiftedChat component

### LINKED ISSUE #1434: GiftedChat cDU comparison does not allow you to update messages
#### Issue Description

In https://github.com/FaridSafi/react-native-gifted-chat/commit/ac984958c3761c15a9a1ebc6c6fca8d49ecd2bec the `componentDidUpdate` comparison for messages was changed, however the comparison no longer allows us to update messages.

e.g.

-> user sends message

```js
// messages array passed to <GiftedChat />
[{ id: 1, status: 'SENDING' }]
```

-> websocket subscription updates message status

```js
// messages array passed to <GiftedChat />
[{ id: 1, status: 'DELIVERED' }]
```

The `componentDidUpdate` only is checking if the array lengths are still the same, which they are, and so it does not update and my `renderBubble` method is not called.

#### Expected Results

The `componentDidUpdate` should either a) check if contents of messages array has changed, or b) allow the user an opt-in like the [shouldMessageUpdate](https://github.com/FaridSafi/react-native-gifted-chat/blob/3cbadaf07ba683d160b35d6a9c42c2204ce48886/src/Message.tsx#L99) function that is passed to `Message` component

#### Additional Information

* Nodejs version: 10.16.0
* React version: 16.9.0
* React Native version: 0.61.x
* react-native-gifted-chat version: 0.10.0
* Platform(s) (iOS, Android, or both?): both
* TypeScript version: 3.6.3

audiolion:
Allow the componentDidUpdate method to be overrided by a custom comparator of prevProps and this.props.

Closes #1434

### DIFF
--- a/src/GiftedChat.tsx
+++ b/src/GiftedChat.tsx
@@ -177,6 +177,11 @@ export interface GiftedChatProps<TMessage extends IMessage = IMessage> {
     props: Message['props'],
     nextProps: Message['props'],
   ): boolean
+  /* Custom comparator to be used in GiftedChat's componentDidUpdate method to override the default memoization implementation */
+  shouldUpdateChat?(
+    prevProps: GiftedChatProps<TMessage>,
+    props: GiftedChatProps<TMessage>,
+  ): boolean
 }
 
 export interface GiftedChatState<TMessage extends IMessage = IMessage> {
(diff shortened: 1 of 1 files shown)


audiolion:
Went with option a) and fixed in https://github.com/FaridSafi/react-native-gifted-chat/commit/e8bf883888ea408d7f90321ee905fa99093aceec

PR #1472, opened 2019-10-16
Fix unwanted scroll to bottom on message change

### LINKED ISSUE #1447: Loading earlier messages should keep ListView at same scroll position
#### Issue Description

After load earlier messages, ListView scroll chat on starting position instead of staying at same scroll position. This bug appear after upgrading from v0.9.11 to v.0.11.0

#### Additional Information

* React Native version: 0.61.1
* react-native-gifted-chat version: v.0.11.0
* Platform(s) (iOS, Android, or both?): both

amiralies:
Closes #1447

### DIFF
--- a/src/GiftedChat.tsx
+++ b/src/GiftedChat.tsx
@@ -407,14 +407,6 @@ class GiftedChat<TMessage extends IMessage = IMessage> extends React.Component<
       this.setMessages(messages || [])
     }
 
-    if (
-      messages &&
-      prevProps.messages &&
-      messages.length !== prevProps.messages.length
-    ) {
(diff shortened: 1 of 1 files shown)


xcarpentier:
when inverted is false

amiralies:
@xcarpentier added that though i think `inverted={false}` is quite unstable and we should make it more stable.
btw can we release this change?

xcarpentier:
yep

adembacajdev:
@FaridSafi we are having again this issue in 2.8.0

Maybe tyou forgot to do the same when you removed componentDidMount and replaced it?

PR #1473, opened 2019-10-16
Add flow-typedefs to npmignore

amiralies:


xcarpentier:
I just published a version including this ;)

PR #1478, opened 2019-10-21
sending giftedChat props to renderChatFooter function prop

anurag-rapidue:
ExtraData property of gifted chat can't be accessed in renderChatFooter. So, Sending giftedChat Props to renderChatfooter prop function

### DIFF
--- a/src/GiftedChat.tsx
+++ b/src/GiftedChat.tsx
@@ -173,6 +173,8 @@ export interface GiftedChatProps<TMessage extends IMessage = IMessage> {
   onQuickReply?(replies: Reply[]): void
   renderQuickReplies?(quickReplies: QuickReplies['props']): React.ReactNode
   renderQuickReplySend?(): React.ReactNode
+  /* Scroll to bottom custom component */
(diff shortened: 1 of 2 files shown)


src/GiftedChat.tsx
```diff
@@ -806,7 +806,7 @@ class GiftedChat<TMessage extends IMessage = IMessage> extends React.Component<
 
   renderChatFooter() {
     if (this.props.renderChatFooter) {
-      return this.props.renderChatFooter()
+      return this.props.renderChatFooter(this.props)
```
xcarpentier: Hi, thanks, can you please add TS definition of this props?

xcarpentier:
any news?

src/GiftedChat.tsx
anurag-rapidue: I am new to TypeScript. However, I feel we may not need the TS definition for this props. If required, where do we need to write it.

anurag-rapidue:
I am new to TypeScript. However, I feel we may not need the TS definition for this props. If required, where do we need to write it.

PR #1508, opened 2019-11-06
Aditional props to send button

rcabarreto:


rcabarreto:
Hello guys. Just to explain what I did here. I need to pass a 'ref' prop to the send button and the composer text field. I could achieve that on the text field using the 'textInputProps' prop. I thought would be nice to have that same logic on the send button. That's basically what I did here. Hope you guys accept it, it's important for me and my team.

Issue #911, opened 2018-06-23
How to customize the accessory height?


ATShiTou:
#### Issue Description

I'd like to develop an app like wechat:

![image](https://user-images.githubusercontent.com/38748693/41807908-32b54c38-7700-11e8-9719-8dee4a68395d.png)

I tried to custom the accessory height like this:
```
<GiftedChat
    renderAccessory={()=><View style={{height:300}}/>
/>
```
But failed.
#### Steps to Reproduce / Code Snippets

I looked up the lib source code ,and try to change accessoryStyle
```
<GiftedChat
    renderAccessory={()=><View style={{flex:1}}/>
    accessoryStyle={{height: 300}}
/>
```
It works, but the accessory has covered the message container.
#### Expected Results

Custom the accessory height and do not cover the message container.

#### Additional Information

* Nodejs version:  10.0.0
* React version: 16.3.1
* React Native version: 0.55.4
* react-native-gifted-chat version: 0.4.3
* Platform(s) (iOS, Android, or both?): [FILL THIS OUT]

moerabaya:
@ATShiTou did you manage to fix it?

ATShiTou:
NO，I give up to use this library and do it myself.

HarryPT:
@ATShiTou Can you share your solution, please?

Durisvk:
@k1115 @moerabaya @ATShiTou 

based on the PR #1524 that is not yet merged,
you can still extend the class both in pure javascript and typescript,

here's a TypeScript example:

```tsx
class GiftedChatExtended extends GiftedChat {
  componentDidUpdate(prevProps: GiftedChatProps, prevState: any, snapshot?: any) {
    super.componentDidUpdate && super.componentDidUpdate(prevProps, prevState, snapshot);

    if (prevProps.minInputToolbarHeight !== this.props.minInputToolbarHeight) {
      this.setState({
        messagesContainerHeight: this.prepareMessagesContainerHeight(
          this.getBasicMessagesContainerHeight() - this.getKeyboardHeight()
        ),
      });
    }
  }
}
```

and then use `GiftedChatExtended` however you'd use original `GiftedChat` component

shamilovtim:
@Durisvk very clever work, nice job.

SiroSong:
why i can't find this.prepareMessagesContainerHeight() in GiftedChat? It isn't here? but I return the value without wrapped with it, the messageContarinerHeight make work yet.

[...]

PR #1524, opened 2019-11-21
adjust message container height with input toolbar/accessory height change

kukiron:
Currently the message container height reamins the same & does not update/adjust with input toolbar/custom accessory height change. The related issue - https://github.com/FaridSafi/react-native-gifted-chat/issues/911.

This PR will update the message container height if the input toolbar height is updated.

### DIFF
--- a/src/GiftedChat.tsx
+++ b/src/GiftedChat.tsx
@@ -421,6 +421,14 @@ class GiftedChat<TMessage extends IMessage = IMessage> extends React.Component<
     if (text !== prevProps.text) {
       this.setTextFromProp(text)
     }
+
+    if (prevProps.minInputToolbarHeight !== this.props.minInputToolbarHeight) {
+      this.setState({
+        messagesContainerHeight: this.prepareMessagesContainerHeight(
+          this.getBasicMessagesContainerHeight(),
+        ),
+      })
+    }
   }
 
   initLocale() {


shamilovtim:
This PR isn't working for me. The accessory height stays the same both with and without the PR. Also toggling the keyboard does not cause a re-render in React, which means this code would never run in that case. So on what render do you expect it to work?

kukiron:
Update your accessroy height in `minInputToolbarHeight` prop which will cause the re-render.
We are currently using this change - 

adjust-accessory-height

---

Update the `minInputToolbarHeight` prop like this - 

```js
minInputToolbarHeight={accessoryOpen ? minHeight + additionalHeight : minHeight}
```

shamilovtim:
Could you provide a full example? I don't understand where accessoryOpen comes from and in my example the keyboard covers the accessory view and the accessory view itself doesn't fully show render. I am testing on iOS.

kukiron:
You can use a local state `accessoryOpen`/`accessoryHeightChanged` & set it to `true` when the accessory is open or height is changed (example image above).

Add the additional height with previously set `minInputToolbarHeight` when the state is `true` & show the full/changed height. Otherwise, show only the initial height. It'll cause the `minInputToolbarHeight` prop change and ,therefore, the component will re-render.

In our case, the accessory is the custom keyboard (in image). A local state `yeppikKeyboardOpen` is initially set to `false`. When the Y-shaped icon is pressed, the state changes to `true` & additional set height (apprx `200px`) of the keyboard is added to the toolbar height.

Hopefully, that helps. You can show some snippets of your implementation, if you still have problem understanding.

Durisvk:
Yes the change in `minInputToolbarHeight` does not rerender the component (it's only applied the first time it renders) which is wrong - thumbs up for this PR

👍

Durisvk:
@kukiron on the other hand when I was trying it locally I had to do:

```tsx
this.setState({
    messagesContainerHeight: this.prepareMessagesContainerHeight(
        this.getBasicMessagesContainerHeight() - this.getKeyboardHeight(),
    ),
});
```

though I'm not aware of all the internals you guys have in this awesome library I just post it here for you to double-check.

my use-case is when the user adds a new line into text input - it'll grow a bit

the simplest example I could make for this case is:

```tsx
const Example = () => {
  const [height, setHeight] = useState(30);
  return (
    <GiftedChat
      renderInputToolbar={() => (
        <TextInput
          multiline
          onContentSizeChange={e => setHeight(e.nativeEvent.contentSize.height)}
        />
      )}
      minInputToolbarHeight={height}
    />
  );
};
```

@xcarpentier I see you already merged, maybe hotfix will be needed.

xcarpentier:
😬 I will revert it then ...

fukemy:
@Durisvk  i got this error: ***this.prepareMessagesContainerHeight is not a function. (In 'this.prepareMessagesContainerHeight(this.getBasicMessagesContainerHeight() - this.getKeyboardHeight())', 'this.prepareMessagesContainerHeight' is undefined)*** plz help

Durisvk:
@fukemy I'm sorry to hear that. I see the `prepareMessagesContainerHeight` method is no longer existing in https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/GiftedChat.tsx .

If you've upgraded your library please be sure to cleanup your local build cache, remove and reinstall node_modules.
If that doesn't please open an issue and provide more details.

fukemy:
Hi @Durisvk , thanks you for fast reply. Can i ask you a question? I want to make a sticker keyboard like the above gif:
https://user-images.githubusercontent.com/17750543/69377506-a476b880-0cd6-11ea-97e0-0bef00b3acaa.gif

 Here i the follow i did: 
 renderaccessory:
 ```
const renderAccessory = (props) => {
        return showSticker ? 
          <View style={{
            width: dimensions.width,
            height: keyboard.keyboardHeight ,
            justifyContent: 'center',
            alignItems: 'center',
            backgroundColor: 'red',
        }}>
            <Text>test</Text>
        </View> : null
        
    }
```

 I want the accessoryView height will equal to keyboard height, here is the keyboard height config:
 ```
import { useKeyboard } from '@react-native-community/hooks'
const keyboard = useKeyboard()

```
 
 then i have a hook to determine when to show stickerKeyboard:
 
 
```

    const [showSticker, setShowSticker] = useState(false)
<GifterChat....
      renderAccessory={renderAccessory}
      accessoryStyle={{height: (showSticker && !keyboard.keyboardShown) ? keyboard.keyboardHeight : 0}}
```

 But i got the problem is i always see the space between keyboard and composer, i dont know why, can u help me to point out the problem? Thanks so much.
  here is the video i captured

https://user-images.githubusercontent.com/8202437/163845685-2a89f915-ca83-41d8-9a42-492656849031.mov

Durisvk:
Hey @fukemy - from the video it really seems there's an extra padding from the bottom since the beginning (even before the keyboard is open).

It's probably to avoid the new iPhone's notches.
There might be a SafeAreaView wrapped around your screen: https://reactnative.dev/docs/safeareaview

> Safe Area's paddings reflect the physical limitation of the screen, such as rounded corners or camera notches (i.e. the sensor housing area on iPhone 13).

But at this point, I am just guessing. You can try to experiment with the following library:

https://www.npmjs.com/package/react-native-iphone-x-helper

and use the `getBottomSpace()` function to calculate the real height:

```ts
      accessoryStyle={{height: (showSticker && !keyboard.keyboardShown) ? keyboard.keyboardHeight - getBottomSpace() : 0}}
```

however, I don't really think this is relevant to this PR since it was reverted therefore it couldn't affect what you're experiencing. So please open a new issue for further discussion.

fukemy:
ok i know the problem, get minInputToolbarHeight return toolbarx2 when renderaccessory

PR #1528, opened 2019-11-27
Add feature viewable messages

matinzd:
The props `onViewableMessagesChanged` added returns array containing list of messages that are visible to user


xcarpentier:
Sorry but really difficult to read the diff because of the other indentation rules you have :/

matinzd:
Ok I will repair it

xcarpentier:
any news?

PR #1533, opened 2019-12-02
Fixing RTL issue.

AZIMAT:
In RTL languages Chats Never Reverse.


xcarpentier:
Why changing the position of the bubble with RTL?

AZIMAT:
@xcarpentier 

On RTL Language Bubbles Place Should Be As Same As LTR that's why we should change position.

xcarpentier:
Sorry but I didn't catch it, do you have any references for this?

PR #1536, opened 2019-12-04
Added textStyle to Message props

nathan-pierotti:


xcarpentier:
OK, but where this style prop is apply?

xcarpentier:
Sorry, this is not relevant.

Issue #1554, opened 2019-12-11
Temporarily disable TextInput


DuranMoodley:
#### 
is there a way to temporarily Disable TextInput for the chat. I only want the user to view the chats but not respond to it. is this possible? Would really appreciate some help with this.

ali5049:
Making a pr for this issue.

ali5049:
@DuranMoodley this issue is fixed. Just pass the disableComposer = {true} prop in GiftedChat to disable the composer.

Hirbod:
@Ali-Haider-Ch ideally, this should also deactivate any CustomAction, I just rendered the CustomActions and passed down the prop

thejogsa:
> @DuranMoodley this issue is fixed. Just pass the disableComposer = {true} prop in GiftedChat to disable the composer.

Perfectly working fine, Thanks man. Saved my day.

jaenjam:
Thank you 😭

PR #1555, opened 2019-12-11
add disableComposer prop

ali5049:
Fix disableComposer Issue, Disabled composer user can only view the chat can't respond.

Issue Reference [Temporarily disable TextInput #1554](https://github.com/FaridSafi/react-native-gifted-chat/issues/1554)

### DIFF
--- a/src/Composer.tsx
+++ b/src/Composer.tsx
@@ -39,6 +39,7 @@ export interface ComposerProps {
   textInputAutoFocus?: boolean
   keyboardAppearance?: TextInputProps['keyboardAppearance']
   multiline?: boolean
+  disableComposer?: boolean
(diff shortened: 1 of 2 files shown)


src/Composer.tsx
```diff
@@ -39,6 +39,7 @@ export interface ComposerProps {
   textInputAutoFocus?: boolean
   keyboardAppearance?: TextInputProps['keyboardAppearance']
   multiline?: boolean
+  disableComposer?: boolean
```
xcarpentier: you have to add it to GiftedChat too

ali5049: Added in GiftedChat. Please review it.

xcarpentier: Thanks.

Issue #1517, opened 2019-11-14
Out-of-the-box Animated Typing Indicator?


JaredDahlke:
First of all thank you for this amazing library! 

I have implemented gifted chat as a pure function of global state. I have also implemented a custom 'renderFooter' where I have an animated typing indicator (looks just like iPhone's typing indicator).
The problem is that I have no way of making it slide up and down to indicate typing, it just appears and dissapears. I'm trying to implement it exactly like Iphone where it slides up and down. It's very nice looking. 

Does gifted chat have this? I am surprised if not because it is a fundamental part of a chat screen that most people would love to be able to easily implement out of the box.   I've looked into react-native-slide-up-panel but cannot implement that since I have gifted chat in a pure function. Any tips or pointers here?

latoda:
Muchas gracias

shamilovtim:
You're saying instead of the element appearing and disappearing you want it to animate in and out?  Have you looked into react-native's Animation API? You could add it and submit a pull request

JaredDahlke:
Can you confirm that there is currently no out of the box typing indicator?
And yes I've already built it in my app using  the Animation API. It looks
pretty good but it has a little bit too much margin on it that I'm having
trouble fixing.

shamilovtim:
Yeah there is only the `renderFooter` and no out of the box typing indicator. You could always submit a pull request with yours if you think it's good and you want other people to be able to use it. In that case yours would be the out of the box option

shamilovtim:
@JaredDahlke Could it be possible to share your work? I'd like to provide something similar to what you did as part of the plugin.

JaredDahlke:
@shamilovtim I just emailed it to you!

Issue #1565, opened 2019-12-17
Feature: Chat typing indicator out of the box


shamilovtim:
#### Issue Description

I propose adding a chat typing indicator out of the box which uses an internal state variable `isTyping` to animate a chat bubble in while `isTyping` is true and to animate the chat bubble out while `isTyping` is false. I'll use vectors and react native animations. Should I have the presence of `renderChatFooter` disable this typing indicator?

If this is desirable and no one is working on it then I'll probably be taking a stab at it next week. 

@xcarpentier

xcarpentier:
Do you mean vector icon?

I think we don’t have to change actual renderChatFooter.

I think it may be interesting to have a new type of message like we already have SystemMessage. Or better this `isTyping` is currently a new sort of SystemMessage.

What do you think?

shamilovtim:
@xcarpentier I was thinking just using vectors themselves to display the ". . ." That way the users can style the dots with whatever color they want.

<img width="141" alt="Screen Shot 2019-12-18 at 2 59 10 PM" src="https://user-images.githubusercontent.com/3495974/71118907-fc55f180-21a6-11ea-8567-f93f52a734b9.png">

 I actually don't need a system message because in my system a system message is simply a `message` with `user: { name: system, avatar: 'system.png'}`. This is an extensible way to have a system message that requires nothing from the plugin's side than what it already offers.  

Example of a system message with system avatar:
<img width="316" alt="Screen Shot 2019-12-18 at 3 03 31 PM" src="https://user-images.githubusercontent.com/3495974/71119177-9ae25280-21a7-11ea-821b-78e928822b63.png">

Regarding `renderChatFooter` it needs to be disabled if the user decides to use the out of the box typing indicator I provide, right?

shamilovtim:
@xcarpentier Just got done. Should have a PR sometime next week if it looks fine.

![Screen Recording 2019-12-28 at 7 42 30 PM](https://user-images.githubusercontent.com/3495974/71551054-ae619a80-29ab-11ea-829c-710fafff89cf.gif)

m4ttheweric:
I am seeing this new prop in the example project, is this working in the main branch? The Accessory bar button in the example setup is supposed to toggle the status if I am understanding correctly, but I am not seeing it. Also, the TS processor believes that "isTyping" is not a <GiftedChat /> prop. I am wondering if there's something still to be finished with this feature...

xcarpentier:
[...]

PR #1581, opened 2019-12-29
Feature: Animated typing indicator

shamilovtim:
Screen Recording 2019-12-28 at 7 42 30 PM

This PR provides an animated typing indicator by default when the user provides a stateful `isTyping` prop. There are no breaking changes here. If the user uses `renderFooter` he will overwrite this typing indicator with his own. The docs have been updated to explain this. All an existing user has to do to upgrade to this indicator is remove their own renderFooter method if they have one.

The indicator transition in both by moving upwards from offscreen and by increasing in height from 0 to its final size. The inside of the indicator animates via a dependency that uses pure vectors.

The example project has been enhanced with a button to trigger the typing indicator.

This closes #1581 #1565


src/MessageContainer.tsx
```diff
@@ -194,14 +198,16 @@ export default class MessageContainer<
     )
   }
 
+  renderTypingIndicator = () => {
```
xcarpentier: Is that possible to provide a way to let user of `renderTypingIndicator` implement it by itself?
Meaning providing a renderTypingIndicator props.
Thanks

xcarpentier:
Awesome !

src/MessageContainer.tsx
shamilovtim: Hey, I can't! The problem is that the animation and chat bubble completely depend on the size of the indicator. There's no option to use flexbox, so it needed to be manually aligned. The easiest way for them to provide their own chat indicator is by using `renderFooter` and plumbing their own indicator, state and animation. This also maintains backwards compatibility with all of the current users who already provided their indicator this way and if I tried to do anything else it would be a breaking change.

README.md
```diff
@@ -271,6 +271,7 @@ interface QuickReplies {
 ## Props
 
 - **`messages`** _(Array)_ - Messages to display
+- **`isTyping`** _(Bool)_ - Typing Indicator state; default `false. If you use`renderFooter` it will override this.
```
xcarpentier: Can you please fix doc here? (ie. `false` 👉 \`)

example-expo/AccessoryBar.js
```diff
@@ -11,11 +11,19 @@ import {
 export default class AccessoryBar extends React.Component {
   render() {
     const { onSend } = this.props
+    const isTyping = this.props.isTyping
```
xcarpentier:
```jsx
const { onSend, isTyping } = this.props
```
?

src/GiftedChat.tsx
```diff
@@ -53,6 +53,8 @@ import QuickReplies from './QuickReplies'
 export interface GiftedChatProps<TMessage extends IMessage = IMessage> {
   /* Messages to display */
   messages?: TMessage[]
+  /* Typing Indicator state */
+  isTyping?: boolean
```
xcarpentier: I think it will be interesting to know who is typing, so we may add more the type, maybe:
```ts
isTyping?: boolean | User[]
```

Even if you don't implement it now :)

src/MessageContainer.tsx
```diff
@@ -361,7 +367,7 @@ export default class MessageContainer<
           : null}
         <FlatList
           ref={this.props.forwardRef}
-          extraData={this.props.extraData}
+          extraData={[this.props.extraData, this.props.isTyping]}
```
xcarpentier: Why?

src/hooks/useUpdateLayoutEffect.ts
```diff
@@ -0,0 +1,22 @@
...
+) {
+  const isInitialMount = useRef(true)
+
+  useLayoutEffect(() => {
+    if (isInitialMount.current) {
+      isInitialMount.current = false
+    } else {
+      effect()
+    }
+  }, dependencies) // eslint-disable-line react-hooks/exhaustive-deps
```
xcarpentier: we use tslint, so eslint disable comment is not relevant

example-expo/AccessoryBar.js
shamilovtim: good catch. I pasted from my own code so I didn't have the onsend prop 😝

src/hooks/useUpdateLayoutEffect.ts
shamilovtim: sorry, this is also a relic from my project, i'll take it out.

src/MessageContainer.tsx
shamilovtim: best practice is for it to be an array. you want me to keep it an obj?

src/GiftedChat.tsx
shamilovtim: How so? You'll never need to know that you yourself are typing. Unless you're imagining group chat?

src/MessageContainer.tsx
shamilovtim: are you asking why isTyping needs to be inside of extraData? it's because the FlatList needs to rerender the footer where the typing animation lives.

src/GiftedChat.tsx
xcarpentier: Yes it's a possibility, but let's keep it as is

src/MessageContainer.tsx
xcarpentier: OK :)

shamilovtim:
Closes #1517

watadarkstar:
Thanks for integrating this typing animation, glad my package is being used for the greater good 🙂🎉

noway:
@shamilovtim What's the point of useUpdateLayoutEffect? It's purposefully breaks the case where I want to show TypingIndicator on mount. I don't understand, why can't your interlocuteur be typing the message already when you open up the chat?!

shamilovtim:
i can't remember perfectly but I believe this was my implementation because we're only listening to typing status when the user is inside the chatroom, so having typing while mounted was not realistic. you're welcome to patch it if it works better for you, it's a very simple change to useLayoutEffect rather than the custom hook.

xcarpentier:
This is published now on v0.14.0! 🎉

fukemy:
hi, any example code? Thanks

fukemy:
istyping not show, code example:
```
        onPressAvatar={() => {
          console.log('longpressavt')
          setIsTyping(true)
        }}
        isTyping={isTyping}
```

PR #1592, opened 2020-01-06
fix keyboardShouldPersistTaps when container is aligned to top

### LINKED ISSUE #1586: keyboardShouldPersistTaps invalid
RN: 0.60.4
react-native-gifted-chat: 0.13.0

I found the props is use in FlatList, but the actual TextInput is not inside FlatList at all. The RN official has clearly informed that the scenario where this property works is when the TextInput is inside the FlatList/ScrollView.

ahce:
Closes #1586 

### Actual behaviour
image

### With fix
But now there is a problem with the typing indicator. 
It can be solved by increasing the [yCoords](https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/TypingIndicator.tsx#L11)
@xcarpentier is there another better solution?

image


xcarpentier:
@shamilovtim Can you provide any thought?

ahce:
> @ahce This PR isn't explained enough for me to understand the problem that it's solving nor is it clear whether this PR is creating a new problem. Is the typing indicator already broken with alignTop or does this PR break the typing indicator?

Solve the behavior of keyboardShouldPersistTaps when container is aligned to top.

For example with keyboardShouldPersistTaps='never', you cannot dismiss the keyboard when touching the container, because the FlatList isn't at full height.

Before, the typing indicator it was hidden where the container wasn't at full height.

Now the typing indicator is visible because it has a [fixed height](https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/TypingIndicator.tsx#L11)

shamilovtim:
In theory the FlatList is only supposed to be as large as its list items, it's unnecessary to make it take up the whole screen just so it's tappable because the react native Keyboard plugin is intended for these types of problems. Changing the FlatList like this completely changes the architecture of the plugin, introduces regressions and changes like the problem with the typing indicator. The plugin has a very large number of configurations and a change like this is a huge can of worms which I think is going to cause a lot of unnecessary extra work. If possible I say just solve this issue with React Native Keyboard.

Issue #1645, opened 2020-02-10
Links in SystemMessages


KDederichs:
#### Issue Description

Hey,
I noticed that if you send a link as system message it will not get parsed and can't be clicked by the user.
Shouldn't a system message be the same as a normal message, except different in style? (Speaking of styles seem to be missing completely)

Is there a way to make them behave more like normal messages would?

#### Steps to Reproduce / Code Snippets

Send a link as a system message

#### Expected Results

The user can click and open the link

#### Additional Information

* Nodejs version: 12.13.1
* React version: 16.9.0
* React Native version:0.61.5
* react-native-gifted-chat version: 0.13.0
* Platform(s) (iOS, Android, or both?): both
* TypeScript version: 3.7.5

KDederichs:
Not stale

KDederichs:
also still not stale

PR #1646, opened 2020-02-11
Use MessageText in SystemMessage

KDederichs:
As outlined in https://github.com/FaridSafi/react-native-gifted-chat/issues/1645 it would be nice to use parsed messages in SystemMessages

### DIFF
--- a/src/TypingIndicator.tsx
+++ b/src/TypingIndicator.tsx
@@ -1,5 +1,6 @@
 import React, { useState } from 'react'
 import { Animated } from 'react-native'
+// @ts-ignore
 import { TypingAnimation } from 'react-native-typing-animation'
 import { useUpdateLayoutEffect } from './hooks/useUpdateLayoutEffect'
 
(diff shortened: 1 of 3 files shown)


KDederichs:
also not stale

KDederichs:
still not stale

src/TypingIndicator.tsx
```diff
@@ -8,9 +9,9 @@ interface Props {
 }
 
 const TypingIndicator = (props: Props) => {
-  const [yCoords, setYCoords] = useState(new Animated.Value(200))
-  const [heightScale, setHeightScale] = useState(new Animated.Value(0))
-  const [marginScale, setmarginScale] = useState(new Animated.Value(0))
+  const [yCoords] = useState(new Animated.Value(200))
```
xcarpentier, line 12: Why?

KDederichs: It's unused, why keep it and let the TS compiler complain about unused values?

Issue #1733, opened 2020-04-10
No audio player shown for audio message.


zholmes1:
#### Issue Description

I am trying to get audio clips to show in my chat, but they are only displaying as an empty bubble.

https://photos.app.goo.gl/hzrgAACxzTwuRhYL6

#### Steps to Reproduce / Code Snippets

This is the IMessage object that is in my IMessage[] messages prop for GiftedChat. audio is a string, and image/text are undefined.

`{
    "_id": 27,
    "audio": "https://s3.us-east-2.amazonaws.com/storage-app-dev/mp3/cb62d4ctnk8uibk1v.mp3",
    "createdAt": 2020-04-10T18:11:40.422Z,
    "image": undefined,
    "text": undefined,
    "user": Object {
      "_id": "2s6aggy7rtjtnp6a8c",
      "avatar": "https://s3.us-east-2.amazonaws.com/storage-app-dev/cb62d4exdk8t4gnhf/thumbnail-26170527_10214525576748161_8444045049145038093_o.jpg",
      "name": "Zach H",
}`

#### Expected Results

Bubble should have an audio player

#### Additional Information

* Nodejs version: 12.16.1
* React version: 16.9.0
* React Native version: 0.61.0
* react-native-gifted-chat version: 0.13.0
* Platform(s) (iOS, Android, or both?): Android
* TypeScript version: 3.7.2

raazatul7:
@zholmes1 Can you help me with the audio thing, it's saying audio is not implemented by Gifted chat

zholmes1:
Gifted chat doesn't have a built in module to play audio. You have to implement it yourself and add a renderMessageAudio prop

raazatul7:
> Gifted chat doesn't have a built in module to play audio. You have to implement it yourself and add a renderMessageAudio prop

Gotcha!
Thanks for replying mate.

0mayur1:
@zholmes1 please send your code

zholmes1:
@elviskankola1 Already answered this sufficiently. All you need to do is create an audio component and supply it as a render prop.to GiftedChat. I just made a play button, expo-av to start playing the sound clip, and change it to a pause button while the clip is playing. Once the clip is done, change it back to a play button. I didn't bother making a seek bar.

Quaid5050:
```
import { Audio } from 'expo-av';

function AudioMessage(props) {
  const [isPlaying, setIsPlaying] = useState(false);
[...]

PR #1747, opened 2020-04-20
renderMessageAudio

RodrigoAngeloValentini:
Add message audio support.


hubermat:
Would fix #1733

ChallengerDefier:
I encountered the same problem, how to solve it?

PR #1838, opened 2020-06-18
MessageObject

danielmejiadev:
- Update the documentation about message object to show developers the way to mark a message with the available states.


sibelius:
I think it would be good to move tests to use https://www.native-testing-library.com/ as it makes sure we can refactor without breaking behavior

try to refactor one component at a time to avoid breaking changes

danielmejiadev:
Thank you very much, Im gonna refactor one component at the time and build unit testing to avoid breaking behavior.

PR #1839, opened 2020-06-18
RefactorBubbleTicks

danielmejiadev:
- Create a folder with bubble component to separte concepts between styles, component and test.
- Refactor the way of display tick in order to make it more readable for developer.

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -1,5 +1,5 @@
 import { MaterialIcons } from '@expo/vector-icons'
-import { AppLoading, Asset, Linking } from 'expo'
+import { AppLoading, Linking } from 'expo'
 import React, { Component } from 'react'
 import { StyleSheet, View, Text, Platform } from 'react-native'
 import { Bubble, GiftedChat, SystemMessage, IMessage, Send } from './src'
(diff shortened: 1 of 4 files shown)


danielmejiadev:
@sibelius @xcarpentier  Could you please review this changes?.
I just wrap the component into a folder and separate the styles, test and component in different files. This is jus to make the files shorter and easy to read.

I also refactored the renderTick in order to create variables with readable names, in this way you know where and why the ticks should be displayed.

xcarpentier:
Sorry I'm not OK. 
@sibelius please do not merge this.

danielmejiadev:
Could you please share with me, What is wrong or what should be changed?

sibelius:
I prefer to keep the styles in the same file that uses it

danielmejiadev:
@sibelius Thanks, I got the idea about how can I contribute. So I rewrite the PR with your comments. Please take a look. Again

xcarpentier:
Hi @danielmejiadev,
Thanks for your work.
Can you please keep indentation also because it's hard to see differences when your changes is just another indentation?
Thanks in advance.

xcarpentier:
@danielmejiadev Should we keep this PR regarding other work you started?
I don't think so, then please remove it ;)

danielmejiadev:
@xcarpentier There is not problem. Im gonna remove it

PR #1856, opened 2020-07-02
[Refactor] Remove unnecessary keyboard listener in empty chat

abdelmagied94:
In case if the chat is empty, two keyboard listeners is registered (One in `MessageContainer` and the other in the rendered `FlatList`). I think we don't need the `MessageContainer` one since the `FlatList` one is already fired even if it is empty.

### DIFF
--- a/src/MessageContainer.tsx
+++ b/src/MessageContainer.tsx
@@ -5,7 +5,6 @@ import {
   FlatList,
   View,
   StyleSheet,
-  Keyboard,
   TouchableOpacity,
   Text,
   ListViewProps,
@@ -143,70 +142,6 @@ export default class MessageContainer<
 
   state = {
     showScrollBottom: false,
-  }
-
-  componentDidMount() {
-    if (this.props.messages && this.props.messages.length === 0) {
-      this.attachKeyboardListeners()
-    }
-  }
-
-  componentWillUnmount() {
-    this.detachKeyboardListeners()
-  }
-
-  componentDidUpdate(prevProps: MessageContainerProps<TMessage>) {
-    if (
-      prevProps.messages &&
-      prevProps.messages.length === 0 &&
-      this.props.messages &&
-      this.props.messages.length > 0
-    ) {
-      this.detachKeyboardListeners()
-    } else if (
-      prevProps.messages &&
-      this.props.messages &&
-      prevProps.messages.length > 0 &&
-      this.props.messages.length === 0
-    ) {
-      this.attachKeyboardListeners()
-    }
-  }
-
-  attachKeyboardListeners = () => {
-    const { invertibleScrollViewProps: invertibleProps } = this.props
-    if (invertibleProps) {
-      Keyboard.addListener(
-        'keyboardWillShow',
-        invertibleProps.onKeyboardWillShow,
-      )
-      Keyboard.addListener('keyboardDidShow', invertibleProps.onKeyboardDidShow)
-      Keyboard.addListener(
-        'keyboardWillHide',
-        invertibleProps.onKeyboardWillHide,
-      )
-      Keyboard.addListener('keyboardDidHide', invertibleProps.onKeyboardDidHide)
-    }
-  }
-
-  detachKeyboardListeners = () => {
-    const { invertibleScrollViewProps: invertibleProps } = this.props
-    Keyboard.removeListener(
-      'keyboardWillShow',
-      invertibleProps.onKeyboardWillShow,
-    )
-    Keyboard.removeListener(
-      'keyboardDidShow',
-      invertibleProps.onKeyboardDidShow,
-    )
-    Keyboard.removeListener(
-      'keyboardWillHide',
-      invertibleProps.onKeyboardWillHide,
-    )
-    Keyboard.removeListener(
-      'keyboardDidHide',
-      invertibleProps.onKeyboardDidHide,
-    )
   }
 
   renderTypingIndicator = () => {


Hirbod:
@abdelmagied94 @xcarpentier I can confirm, that this PR doesn't break anything for me. My tests were just fine and I can also see that the height calculations do not jump nor stutter anymore.

Maybe some clarification what the second listener was good for?

TL;DR: LGTM

abdelmagied94:
> Maybe some clarification what the second listener was good for?

@Hirbod I don't figure out why it was needed. It is just attached if the chat is empty. When the length of messages become greater than zero, it is detached. I thought that the FlatList doesn't trigger keyboard events if it's empty but I found it the events are always triggered from it.

Hirbod:
I also tested this PR against empty chats, full chats, with messages, deletes etc. Nothing breaks, everything remains the same and the keyboard opens faster and sets its position better then before. I think its safe but lets see what Xaviers thoughts are.

abdelmagied94:
@Hirbod @xcarpentier I think I figured out why it was introduced. I found this change made in this PR [#898](https://github.com/FaridSafi/react-native-gifted-chat/pull/898). If you look at the message container `render` method when this PR made, you will find a condition if the messages is empty. Now, this condition is removed and we use `ListEmptyComponent` so the `FlatList` is always rendered.

xcarpentier:
Hi @abdelmagied94,

I'm seeing an issue on the `keyboardShouldPersistTaps={'never'}`.

Steps to reproduce:
1. No messages in the list (also no `loadEarlier`).
2. Tap on Composer (ie. TextInput).
3. See the Keyboard appear.
4. Tap on the Empty List, nothing happens.
5. Type a message and send, now that's works.

Do you think you can fix it?

xcarpentier:
Hi @doomsower,
Any thought?

Hirbod:
I also use the ListEmptyComponent, thats maybe the reason why I do not have any issues

xcarpentier:
@Hirbod Are you using `renderChatEmpty` or from `listProps`?
Can you share the code of this empty component?
Event if the `renderChatEmpty` is not fill we return a View with flex style
https://github.com/FaridSafi/react-native-gifted-chat/blob/2d9232fb234c12c2074386bc2c0f94dfaafd2010/src/MessageContainer.tsx#L321-L333

Hirbod:
Looks like I had something wrong at top of my head.
This is how my Component looks like. (I didn't refactor to FC currently, doing this when RNGC has finished to do so)

```ts
<View style={{ flex: 1, backgroundColor: this.props.theme.colors.surface }}>
        <GiftedChat
          messagesContainerStyle={{ backgroundColor: this.props.theme.colors.background }}
          // @ts-ignore
          ref={this.giftedChatRef}
          renderMessageImage={this.renderMessageImage}
          renderSend={this.renderSend}
          locale="de"
          bottomOffset={this.isIphoneX() ? 22 : -12}
          scrollToBottom
          messages={state.messages}
          onSend={this.onSend}
          loadEarlier={state.loadEarlier}
          onLoadEarlier={this.onLoadEarlier}
          isLoadingEarlier={state.isLoadingEarlier}
          placeholder={
            !state.isConnectedToSocket || !state.dataHasBeenLoaded ? "Verbinde..." : "Aa"
          }
          isTyping={this.state.isTyping}
          timeFormat="HH:mm"
          label="Senden"
          minComposerHeight={28}
          maxComposerHeight={106}
          renderComposer={this.renderComposer}
          onLongPress={this.onLongPress}
          renderInputToolbar={this.renderInputToolbar.bind(this)}
          renderLoadEarlier={this.renderLoadEarlier}
          inverted
          minInputToolbarHeight={50}
          onPressAvatar={this.onAvatarPress}
          onInputTextChanged={this.inputTextChanged}
          disableComposer={!state.isConnectedToSocket || !state.dataHasBeenLoaded}
          user={{
            _id: me.id, // sent messages should have same user._id
          }}
          scrollToBottomComponent={this.renderScrollToBottom}
          scrollToBottomStyle={styles.scrollToBottomStyle}
          renderActions={this.renderCustomActions}
          renderBubble={this.renderBubble}
          renderSystemMessage={this.renderSystemMessage}
          listViewProps={{
            alwaysBounceVertical: false,
            directionalLockEnabled: true,
            removeClippedSubviews: false,
            initialNumToRender: 12,
            contentContainerStyle: {
              backgroundColor: this.props.theme.colors.background,
              paddingTop: 15,
              paddingBottom: 5,
            },
          }}
          shouldUpdateMessage={() => {
            return false;
          }}
        />
      </View>
```

abdelmagied94:
Hi @xcarpentier,

Please, Could you tell me in which branch this issue happens? I tried the repo example with empty messages on `master` and this branch and it works well on Android and iPhone? I test it on iPhone 11 Pro (iOS 13.5) and Android 10.

xcarpentier:
No worries it's not related to this PR!

Actually this is apparently another issue also present on master.

👉  [Video](https://www.dropbox.com/s/8impa23i6la3n66/keypersist.mp4?dl=0&raw=1)

1. Tap on `Composer` (focus composer)
2. Tap on InputToolbar or other things like image picker - note the Composer keep the focus
3. Select image
4. Tap on List and nothing happens and Composer still focus
(Need to force the blur of TextInput?) 

Then I will accept this PR

abdelmagied94:
> Actually this is apparently another issue also present on master.

@xcarpentier I tried to resolve this issue in this PR #1858. Please, check it.

PR #1858, opened 2020-07-02
[Fix] Keyboard persist taps issue after selecting image

abdelmagied94:
The issue happens during opening image picker and the text input is already focused. In this case, the text input lost its focus after opening the picker so when the picker closed we can't dismiss the keyboard by tapping on container so we need to re-focus the text input. The steps to produce the issue are in this [Video](https://www.dropbox.com/s/8impa23i6la3n66/keypersist.mp4?dl=0&raw=1) (Thanks to @xcarpentier)

### DIFF
--- a/src/GiftedChat.tsx
+++ b/src/GiftedChat.tsx
@@ -391,7 +391,7 @@ class GiftedChat<TMessage extends IMessage = IMessage> extends React.Component<
   invertibleScrollViewProps: any = undefined
   _actionSheetRef: any = undefined
   _messageContainerRef?: RefObject<FlatList<IMessage>> = React.createRef()
-  _isTextInputWasFocused: boolean = false // Indicator to show if the keyboard was focused before hiding keyboard or not
+  _isTextInputWasFocused: boolean = false
   textInput?: any
 
   state = {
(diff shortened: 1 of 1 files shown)


src/GiftedChat.tsx
```diff
@@ -391,6 +391,7 @@ class GiftedChat<TMessage extends IMessage = IMessage> extends React.Component<
   invertibleScrollViewProps: any = undefined
   _actionSheetRef: any = undefined
   _messageContainerRef?: RefObject<FlatList<IMessage>> = React.createRef()
+  _isTextInputWasFocused: boolean = false // Indicator to show if the keyboard was focused before hiding keyboard or not
```
xcarpentier, line 394: Can you please remove comment, I think the name is already explicit? thanks

src/GiftedChat.tsx
```diff
@@ -588,6 +589,19 @@ class GiftedChat<TMessage extends IMessage = IMessage> extends React.Component<
   }
 
   onKeyboardWillShow = (e: any) => {
+    // Focus the text input only if it was focused before showing keyboard
```
xcarpentier, line 592: Can you please extract this code into a function of that class with an explicit name (and remove comments)?
Thanks

src/GiftedChat.tsx
```diff
@@ -602,6 +616,13 @@ class GiftedChat<TMessage extends IMessage = IMessage> extends React.Component<
   }
 
   onKeyboardWillHide = (_e: any) => {
+    // Use condition since the `onKeyboardWillHide` may be called
```
xcarpentier, line 619: Can you please also extract this code into a function of that class with an explicit name (and remove comments)?
Thanks

xcarpentier:
Many thanks 👍

PR #1897, opened 2020-08-13
Add textProps to Day component

gabrieldonadel:
This PR adds the ability of us passing specific props directly to the `Text` component used on `Day.tsx`. This is especially useful when we need to use something as `maxFontSizeMultiplier` for example.

### DIFF
--- a/src/Day.tsx
+++ b/src/Day.tsx
@@ -7,6 +7,7 @@ import {
   StyleProp,
   ViewStyle,
   TextStyle,
+  TextProps,
 } from 'react-native'
 import dayjs from 'dayjs'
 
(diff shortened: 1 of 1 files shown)


xcarpentier:
Hi @gabrieldonadel 
Thanks!
Can you also execute a test and verify that not change the snapshot of this component?
Thanks in advance

gabrieldonadel:
> Hi @gabrieldonadel
> Thanks!
> Can you also execute a test and verify that not change the snapshot of this component?
> Thanks in advance

Just checked, it does not affect the component snapshot.

Here are the test results:
`yarn test`
image
`yarn test:coverage`
image
`yarn test:watch`
image

gabrieldonadel:
@xcarpentier Do you mind merging this?

PR #1941, opened 2020-10-13
add renderFooter prop to Bubble

sonnyp:
I didn't want to increase API surface too much so it's not exposed as a prop on GiftedChat but happy to if you think it's preferable.
`renderBubbleFooter`

### DIFF
--- a/src/Bubble.tsx
+++ b/src/Bubble.tsx
@@ -161,6 +161,7 @@ export interface BubbleProps<TMessage extends IMessage> {
   renderUsername?(): React.ReactNode
   renderQuickReplySend?(): React.ReactNode
   renderQuickReplies?(quickReplies: QuickReplies['props']): React.ReactNode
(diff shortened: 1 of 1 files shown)


src/Bubble.tsx
```diff
@@ -319,6 +320,17 @@ export default class Bubble<
     return null
   }
 
+  renderFooter() {
+    const {
+      currentMessage,
+      renderFooter
+    } = this.props
+    if (renderFooter && currentMessage) {
```
FauricioValencia, line 328: what do you think if do this:

```
renderFooter() {
    const {
      currentMessage,
      renderFooter
    } = this.props

   if (renderFooter && currentMessage) return renderFooter(currentMessage)
   return null
}
```

sonnyp: No logical difference and my version is using gifted chat coding style.

PR #1961, opened 2020-11-19
adds textInputStyle

reggi:


tkow:
@xcarpentier 
This is valid fix. Now using Typescript, textInputStyle prop warn type error though it can be used.
Can you check when you have time?

PR #1987, opened 2021-01-20
Use onLayout instead of onContentSizeChange for composer height

mtieltjes:
I had some issues where the composer height was incorrectly calculated. During debugging I found it to be an issue with the React Native `onContentSizeChange` API.

Turns out that `onContentSizeChange` is not all that accurate. For example, if you fill a line with 'M's, and the last m gets pushed onto the next line, `onContentSizeChange` will still trigger a height and width for it being on the same line.
The exact offset differs, on some screen dimensions I didnt have this problem initially, but when padding is applied to the `TextInput` is occurs on every instance.

This result in the InputToolbar not re-calibrating correctly, and slowly moving beneath the keyboard the longer you message gets.

Demo of the issue, with the registered `contentSize` logged below.

```js
styleInputProps={{ style: {
    borderWidth: 2,
    borderColor: '#949393',
    backgroundColor: 'white',
    flex: 1,
    marginRight: 5,
    minHeight: 40,
    paddingHorizontal: 20,
    paddingVertical: 0,
}}}
```

https://user-images.githubusercontent.com/9537966/105214554-7302b280-5b50-11eb-8045-6cb482da9426.mov

This fix changes the API to use the [`onLayout`](https://reactnative.dev/docs/view#onlayout) API of ReactNative, showing more accurate dimensions of the actual component, not the content.

Same environment, but with `onLayout`:

https://user-images.githubusercontent.com/9537966/105215183-3daa9480-5b51-11eb-862f-ffafdda6cbf3.mov


heyalexchoi:
I find that `onLayout` doesn't call when it needs to and so composer resizing breaks.

Issue #1928, opened 2020-09-23
Flatlist - onEndReached fails to trigger, because "distanceFromEnd" has negative value


RosenTomov:
#### Issue Description

Infinite scrolling fails to trigger most of the time on iOS and rarely on Android, because the `distanceFromEnd` returned to the Flatlist's `onEndReached` function is a negative number. You have to scroll slowly to trigger `onEndReached`.

https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/MessageContainer.tsx#L319

`distanceFromEnd > 0` is the culprit. To go around this, we've patched the package to remove that check.

#### Steps to Reproduce / Code Snippets

Enable infinite scrolling and scroll up (`inverted` by default).

#### Expected Results

To trigger `onEndReached`, when the end of the screen is reached.

#### Additional Information

* Nodejs version: 12.18.3
* React version: 16.13.1
* React Native version: 0.63.2
* react-native-gifted-chat version: react-native-gifted-chat
* Platform(s) (iOS, Android, or both?): 
Tested on both:
  - iOS: 
    - v13.6 simulator & device 
  - Android: 
    - v10 emulator 
    - v9 device
* TypeScript version: none

doelgonzo:
This appears to still be occurring. Seems odd to have that check in there at all. Is there a reason around it?

RosenTomov:
Looks like this was marked as "wontfix", here's what you'll need to patch it yourself:
- "patch-package" from npm
- See the set-up and usage section in their docs (tl;dr edit the package's code in node_modules, run a command with it's name and it will generate a patch with the changes).

Here's the file it generated for me in the `patches` folder (in root), you can try using it directly
filename: `react-native-gifted-chat+0.16.3.patch`
```
diff --git a/node_modules/react-native-gifted-chat/lib/MessageContainer.js b/node_modules/react-native-gifted-chat/lib/MessageContainer.js
index 193772a..004eca7 100644
--- a/node_modules/react-native-gifted-chat/lib/MessageContainer.js
+++ b/node_modules/react-native-gifted-chat/lib/MessageContainer.js
@@ -172,7 +172,6 @@ export default class MessageContainer extends React.PureComponent {
         this.onEndReached = ({ distanceFromEnd }) => {
             const { loadEarlier, onLoadEarlier, infiniteScroll, isLoadingEarlier, } = this.props;
             if (infiniteScroll &&
-                distanceFromEnd > 0 &&
[...]

PR #2010, opened 2021-03-04
* fix infiniteScroll issue #1928

Iskander508:
Fix for `infiniteScroll` issue #1928

Run the `onLoadEarlier` even when hitting the edge, but not when initially opening the chat (user scroll has to happen first)

### DIFF
--- a/src/MessageContainer.tsx
+++ b/src/MessageContainer.tsx
@@ -142,6 +142,7 @@ export default class MessageContainer<
 
   state = {
     showScrollBottom: false,
+    hasScrolled: false,
   }
 
   renderTypingIndicator = () => {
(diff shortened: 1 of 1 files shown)


xcarpentier:
Can you provide a gif?

Iskander508:
Yes, before change:
before

after change:
after

efstathiosntonas:
@Iskander508 there's a typing missing:

current: 
```
interface State {
  showScrollBottom: boolean,
}
```

should be:
```
interface State {
  showScrollBottom: boolean,
  hasScrolled: boolean
}

```

efstathiosntonas:
Thanks @Iskander508 !

Issue #1990, opened 2021-01-25
GiftedChat.d.ts does not have the correct typing for renderAvatar


jkepps:
#### Issue Description
The type definition file for GiftedChat does not have the correct typing for the `renderAvatar` prop.
```ts
renderAvatar?(props: Avatar<TMessage>['props']): React.ReactNode;
```
`null` is not allowed, and as a result, a typescript error is thrown when passing `null` as the value.

#### Steps to Reproduce / Code Snippets
```jsx 
<GiftedChat renderAvatar={null} />
```

#### Expected Results
There is no typescript error.

#### Additional Information

* Nodejs version: 14.6.0
* React version: 16.13.1
* React Native version: 0.63.4
* react-native-gifted-chat version: 0.16.3
* Platform(s) (iOS, Android, or both?): both
* TypeScript version: 4.0.2

efstathiosntonas:
Fixed at https://github.com/FaridSafi/react-native-gifted-chat/pull/2023

PR #2023, opened 2021-03-26
Fixed type definition to allow `null` as acceptable value for `render…

devautor:
It is documented that

  > `renderAvatar` (Function) - Custom message avatar; set to `null` to not render any avatar for the message

  Since `renderAvatar={null}` used to show type error and so I had used `renderAvatar={() => null}` instead thinking that it'd reflect the same behavior.
  As I got back to the code a week later, the bad trip began with this:

rn-gifted-chat-bug

  1. In `Message.tsx`, `this.renderAvatar()` is called. Since user's avatar is acceptable to be an `undefined` value, it proceeds ahead by calling `return <Avatar {...props} />` for other user's messages.

  2. After `renderAvatar` is called, in the `render()` function of `Avatar`, this check causes all the trouble:
  `if (renderAvatar === null) { return null;}`

  3. Thereafter, `GiftedAvatar` is rendered with default `null` values for all these messages except the last one. Note that when avatar is `undefined`, `renderInitials` is used in `GiftedAvatar` to show avatar name initials and hence `undefined` is an expected value here. That's why `renderAvatar` prop for `GiftedChat` must be defined.

  4. For the last message by the other use in this sequence of messages, `renderAvatar()` function is called in `Avatar.js` itself, where my function `this.props.renderAvatar` being present renders a `null` React element.

  What worked for me: **Fixed type definitions for `renderAvatar` to allow `null` as an accepted value.**

### DIFF
--- a/src/Avatar.tsx
+++ b/src/Avatar.tsx
@@ -52,7 +52,7 @@ export interface AvatarProps<TMessage extends IMessage> {
   imageStyle?: LeftRightStyle<ImageStyle>
   containerStyle?: LeftRightStyle<ViewStyle>
   textStyle?: TextStyle
-  renderAvatar?(props: Omit<AvatarProps<TMessage>, 'renderAvatar'>): ReactNode
+  renderAvatar?: ((props: Omit<AvatarProps<TMessage>, 'renderAvatar'>) => ReactNode) | null
   onPressAvatar?(user: User): void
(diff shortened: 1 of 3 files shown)


efstathiosntonas:
Fixes https://github.com/FaridSafi/react-native-gifted-chat/issues/1990

PR #2025, opened 2021-03-29
Fix scroll to bottom for screen readers

jlapstra:
# Scroll To Bottom Fix for Screen Readers

1. Fixes an issue where the scroll to bottom button was not clickable when the component overlapped with another element such as the chat bubble. 
2. Made the expo test app elements accessible by removing accessible from the wrapped View (which caused everything tapped with voiceover to read "main".

# Explanation

While debugging I came across : [This stack overflow question](https://stackoverflow.com/questions/36938742/touchablehighlight-not-clickable-if-position-absolute/51666282) with the answer being to swap the components order. 

# Before

Image from iOS

# Result
This fixes VoiceOver for iOS users and TalkBack for Android users, such that they can successfully tap the scroll to bottom button when it overlaps other content, and be taken to the bottom of the message list.

Image from iOS (1)


mariomurrent-softwaresolutions:
So no updates on this one, right?

Johan-dutoit:
If you're able to resolve the conflicts, then this can be merged in.

jlapstra:
@Johan-dutoit Sorry about that! Did not get a notification. This PR should be good to go now!

PR #2058, opened 2021-05-18
Fixed the empty background cannot scroll problem

penghouho:


cutenesss:
@penghouho have this been merged?

penghouho:
@cutenesss I think it is not merged, the main change is wrapped the View in TouchableWithoutFeedback so that the view cannot be scrolled in file src/Message.tsx, thanks.

cutenesss:
image
thank you for the fixing. so this is the part that we need to fix this?

penghouho:
@cutenesss As far as I know, yes, but maybe you need to test it on your side first to ensure it, because it is change introduced one year ago. thanks.

PR #2073, opened 2021-07-17
fix crash action sheet

anhdevit:
I want to validate actionSheetRef because it make some crash in my application please check and merge


anhdevit:
@FaridSafi Hi, can you check this MR?

PR #2100, opened 2021-09-10
Add QuickReplyTextStyle

harry2507527:
Currently now it is not possible to style the text within. It is now only possible to style touchable opacity but not the text inside.


Johan-dutoit:
If you're able to resolve the conflicts, then this can be merged in.

kesha-antonov:
Not an issue anymore https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/QuickReplies.tsx#L54

PR #2217, opened 2022-06-10
Fix: QuickReplies component is returning null before a react hook & breaking rules of hooks.

### LINKED ISSUE #2216: QuickReplies component is returning null when shouldComponentDisplay = false & breaking rules of hooks. Clicking on QuickReplies is throwing an error
#### Issue Description
- QuickReplies component is returning null when shouldComponentDisplay = false & breaking rules of hooks.
- The handlePress method is using useCallbackOne after ` if (!shouldComponentDisplay)  { return null;  }` code.
- So, that function does not work properly.
- Clicking on a quick-reply throws the following error:

```
Uncaught Error: Rendered fewer hooks than expected. This may be caused by an accidental early return statement.
```

- Use the `react-native-gifted-chat/lib/QuickReplies` in your app.
- Render some message that does not have a quick reply. The QuickReplies component is hidden initially as `shouldComponentDisplay is false`.
- Then render another message that does have a quick reply.
- Click on the quick reply.
- Error is thrown.

#### Expected Results
- Should not throw an error on clicking the quick replies.

#### Additional Information

* Nodejs version: v16.15.0
* React version: 17.0.1
* React Native version: 0.64.3
* react-native-gifted-chat version: 1.0.1
* Platform(s) (iOS, Android, or both?): web (Firefox 101.0.1)
* TypeScript version: 4.7.2

rajatkantinandi:
- Return null when !shouldComponentDisplay after the useCallbackOne hook
- Closes #2216

### DIFF
--- a/src/QuickReplies.tsx
+++ b/src/QuickReplies.tsx
@@ -91,10 +91,6 @@ export function QuickReplies({
     return false
   }, [currentMessage, nextMessage])
 
-  if (!shouldComponentDisplay) {
-    return null
-  }
-
   const handlePress = useCallbackOne(
     (reply: Reply) => () => {
       if (currentMessage) {
@@ -129,6 +125,10 @@ export function QuickReplies({
         messageId: currentMessage!._id,
       })),
     )
+  }
+
+  if (!shouldComponentDisplay) {
+    return null
   }
 
   return (


admehta01:
@Johan-dutoit any chance this can be merged in? Blocking use of v1 for us.

Johan-dutoit:
Released- https://github.com/FaridSafi/react-native-gifted-chat/releases/tag/v1.0.2

Johan-dutoit:
@admehta01  Released- https://github.com/FaridSafi/react-native-gifted-chat/releases/tag/v1.0.2

PR #2218, opened 2022-06-10
fix(Composer): fix typo preventing onInputSizeChanged from being called

amerikan:
Somehow it seems during a refactor the wrong properties were used.

https://github.com/FaridSafi/react-native-gifted-chat/blame/260902effd61b8cbf8b04aa8e4f3bb817f5f970d/src/Composer.tsx#L92-L93

PR #2219, opened 2022-06-12
Problem with minComposerHeight.

adamgajzlerowicz:
The minComposerHeight prop is not working correctly for me. Seems that having the keyboard hidden/shown causes the message list to be placed below the message container.
Also, event with multiline prop I am not able to scroll more than 1 line of message at the time.
I started looking at the code, and seems that the current approach to keyboard hiding is using absolute and relative positioning. I was wondering, if perhaps it would be cleaner to use flex based containers. This would allow for the input to grow as needed. It also works correctly with the prop now supplied.

Please let me know what you think

Simulator Screen Recording - iPhone 13 Pro - 2022-06-12 at 16 34 42

### DIFF
--- a/example/App.tsx
+++ b/example/App.tsx
@@ -23,6 +23,8 @@ const styles = StyleSheet.create({
   container: { flex: 1 },
 })
 
+const findStep = step => message => message._id === step
+
 const user = {
   _id: 1,
   name: 'Developer',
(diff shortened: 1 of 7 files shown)


Johan-dutoit:
Any chance this fix solves it for you? https://github.com/FaridSafi/react-native-gifted-chat/pull/2218

adamgajzlerowicz:
It doesn't really fix it.
I did checkout the latest version from master here and seems if I run it on Ios emulator, the composer multiline also shows one line only.

Unless that is expected, but for me personally it makes it hard to write longer messages (scrolling with fingers is also hard).

Is there a reason flex layout was not used here in a first place? If you agree we should/can change this, I am happy to re-do the pull request

adamgajzlerowicz:
I combined fix above with `minComposerHeight`. That fixed it for me.

PR #2227, opened 2022-06-29
Update Message.tsx

Hassan-Riaz:
Render Day functionality should be generic as when we renderDay handler is provided from front end then this logic is duplicated


Johan-dutoit (approved):
Seems sensible.   Can you share an example of the before and after?

PR #2264, opened 2022-08-15
support for static local image resources added

mujahidali88094:


mujahidali88094:
Image property inside a message now can take local static resource as value. Before only strings (links to a remote images) were acceptable.

PR #2276, opened 2022-09-01
upgrade react-native-safe-area-context dependency to fix issue with expo SDK 46

kyoto21:
fix react-native-safe-area-context dependency to same as the required of expo SDK 46 from ^4.2.4 to ^4.3.1


kyoto21:
to fix issue "Invariant Violation: Tried to register two views with the same name RNCSafeAreaProvider, js engine: hermes" when using expo SDK 46

PR #2283, opened 2022-09-12
chore(lint): adds stricter eslint configs

doublethefish:
There is no functional change here, but it does prepare the lint to support future work.

I noticed several 'obvious' issues were not being picked up by linting.

To help get to solving those problems I have extended the linting for react-native, supporting jest, prettier, jsx and using the airbnb defaults.

The issues that should probably be considered are marked as 'FIXME' in the eslint config. There is at least one of the issues listed there occurring and I have made no qualitative assessment of whether they should be kept off or not. I have turned errors off for gifted-chat that do not makes sense for the library like `no-underscore-dangle` and `@typescript-eslint/no-explicit-any`.


kesha-antonov:

PR #2285, opened 2022-09-18
Added External Ref For Accesing Message Flatlist

rahulrajsbkk:
Here I Have Added FlatList Ref Accessible From Props To Make The Chat Scroll To Specific Message


Johan-dutoit:
This has been implemented in the latest 2.3.0

PR #2286, opened 2022-10-03
Bump react-native-safe-area-context version

rmdevtorres:
Bump react-native-safe-area-context version to 4.4.1


Johan-dutoit:
Should be fixed with https://github.com/FaridSafi/react-native-gifted-chat/pull/2353 and https://github.com/FaridSafi/react-native-gifted-chat/releases/tag/v2.0.0

PR #2288, opened 2022-10-08
Add `renderAboveMessages` prop

Myzel394:
Sometimes you want to render something above the messages (Signal for example shows the avatar at the very beginning of a conversation). This is now possible with this PR.


Myzel394:
@FaridSafi Could you take a look at this?

murph:
I think it might make sense to render this within the MessageContainer (so it's not always on screen, but you can scroll away from it) - if the intent is to more closely match something like Signal's style.

You can do this already by overloading "renderLoadEarlier" but that's not exactly clean / would conflict if you were already using that functionality.

kesha-antonov:
Hi
You can also use listProps.ListHeaderComponent
Did you try that?

Screenshot 2025-11-17 at 00 05 21

PR #2313, opened 2022-12-07
add animation on send message

mmsmadi:
This PR includes bubble animate on send message


mmsmadi:
Animate bubble on send

fukemy:
calling `LayoutAnimation` inside `componentDidUpdate` will cause much glitched issue

Issue #2274, opened 2022-08-25
Getting extra space on left side chat messages.


rohank45:
![image](https://user-images.githubusercontent.com/77183753/186664302-76f7c14b-4808-46b3-8685-cac5ba8663ad.png)

rohankurane45:
> did u modified the example? It's seem the render avatar cause problem

<GiftedChat
            // alwaysShowSend
            isAnimated
            scrollToBottom={true}
            inverted={false}
            // infiniteScroll={true}
            // isKeyboardInternallyHandled={false}
            // keyboardShouldPersistTaps="never"
            // bottomOffset={0}
            user={{
              _id: senderId,
            }}
            textInputProps={{
              autoCorrect: false,
            }}
            listViewProps={{
              // nestedScrollEnabled: true,
              contentContainerStyle: {
                flex: 1,
                justifyContent: 'flex-end',
                marginBottom: 10,
              },
            }}
            text={allocated_sellers?.toString().replace(/,/g, '')}
            messages={messages.sort(
              (a, b) => Date.parse(a.createdAt) - Date.parse(b.createdAt),
            )}
            onLongPress={onLongPress}
            onSend={messages => handleSendMsg(messages)}
            onDelete={messages => handleDeleteMsg(messages)}
            renderAvatar={() => null}
            renderBubble={renderBubble}
            renderSend={renderSend}
            renderInputToolbar={props => customtInputToolbar(props)}
          />

rohankurane45:
> did u modified the example? It's seem the render avatar cause problem

<GiftedChat
// alwaysShowSend
isAnimated
scrollToBottom={true}
inverted={false}
// infiniteScroll={true}
// isKeyboardInternallyHandled={false}
// keyboardShouldPersistTaps="never"
// bottomOffset={0}
user={{
_id: senderId,
}}
textInputProps={{
autoCorrect: false,
}}
listViewProps={{
// nestedScrollEnabled: true,
contentContainerStyle: {
flex: 1,
justifyContent: 'flex-end',
marginBottom: 10,
},
}}
text={allocated_sellers?.toString().replace(/,/g, '')}
messages={messages.sort(
(a, b) => Date.parse(a.createdAt) - Date.parse(b.createdAt),
)}
onLongPress={onLongPress}
onSend={messages => handleSendMsg(messages)}
onDelete={messages => handleDeleteMsg(messages)}
renderAvatar={() => null}
renderBubble={renderBubble}
[...]

PR #2327, opened 2023-01-04
Fixes null type for renderAvatar prop

lsthornt:
The `renderAvatar` prop must accept `null` as value to prevent issues like #2274.

`null` was included in the union type of the `renderAvatar` return value, and thus was not caught by [this TS rule](https://github.com/microsoft/TypeScript/pull/39570)


lsthornt:
Is it possible to re-run/remove the CI check? Looks like newer PRs are not facing this issue

mendrixsebas (approved):

PR #2330, opened 2023-01-09
Add `imageSourceProps` to `MessageImage`

amhinson:
There is currently no way to add additional headers to the image shown in `MessageImage`. This can be an issue if you need to add credentials to the header when making a request, for instance.

Usage: 

```tsx
<MessageImage
    {...props}
    imageSourceProps={{
        headers: {
            Authorization: `Bearer ${authToken}`
        }
    }}
/>
```

### DIFF
--- a/src/MessageImage.tsx
+++ b/src/MessageImage.tsx
@@ -62,7 +62,7 @@ export function MessageImage<TMessage extends IMessage = IMessage>({
         <Image
           {...imageProps}
           style={[styles.image, imageStyle]}
-          source={{ uri: currentMessage.image, ...imageSourceProps }}
+          source={{ ...imageSourceProps, uri: currentMessage.image }}
         />
       </Lightbox>
     </View>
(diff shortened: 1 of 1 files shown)


src/MessageImage.tsx
```diff
@@ -59,7 +62,7 @@ export function MessageImage<TMessage extends IMessage = IMessage>({
         <Image
           {...imageProps}
           style={[styles.image, imageStyle]}
-          source={{ uri: currentMessage.image }}
+          source={{ uri: currentMessage.image, ...imageSourceProps }}
```
Johan-dutoit, line 65: Would the following change make more sense?  As it'll preserve the uri coming from the message and not allowed to be overridden. 

```suggestion
          source={{ ...imageSourceProps, uri: currentMessage.image }}
```

amhinson: Yep, updated!

gajjardarshithasmukhbhai (approved):
Hi @amhinson, it's looking good to me. but we will wait for @Johan-dutoit.

Johan-dutoit (approved):

PR #2332, opened 2023-01-17
Updated to use-memo-one for supporting react 18

AlbertoMeQ:
Bumped `use-memo-one` to 1.1.3, which adds support for React 18


Johan-dutoit:
Will be fixed in 3.3.0

PR #2336, opened 2023-01-20
Add wrapInSafeArea prop to README.md

se1exin:
I use gifted chat in a component that already handles safe areas, after about an hour of trying to figure out why gifted chat had a strange space under the input bar I went digging through the code and found the `wrapInSafeArea` prop.

I thought I'd add it to the readme to help others like me in the future :)


amerikan:
No longer will be needed, as built-in SafeArea has been dropped in v2.0.0.

PR #2340, opened 2023-02-01
fix: ⚡ limit rerenders

matthieuMay:
Two modifications : 
=> Do not use setState to manage messages as it implies 2 renders instead of 1
=> Manage extraData so that every MessageContainer render does not create a full re-render of the flatlist

Tested it on a react-native and react-native-web application applying a patch-package


Johan-dutoit:
Some conflicts, but would be happy to get this merged.

PR #2341, opened 2023-02-08
Patch for android workaround

thebigredgeek:


Johan-dutoit:
The PR is bloated with the build artefacts.  Also this is a `React Native` problem and not a gifted chat problem.

PR #2345, opened 2023-02-22
Update MessageImage.tsx

ZhengKehang:
i need use local image


fukemy:
I didnt see which changed?

PR #2234, opened 2022-07-06
Fixed react-native-safe-area-context Dependency

monahans:
Fixed issue raised here: https://github.com/FaridSafi/react-native-gifted-chat/issues/2223

PR #2348, opened 2023-02-26
Fixed react-native-safe-area-context Dependency (again)

JotaMiller:
[This PR](https://github.com/FaridSafi/react-native-gifted-chat/pull/2234) added a solution but i think it lost for some reason.

pre fix:
image

post fix:
image


Johan-dutoit:
Should be fixed with https://github.com/FaridSafi/react-native-gifted-chat/pull/2353 and https://github.com/FaridSafi/react-native-gifted-chat/releases/tag/v2.0.0

Issue #2352, opened 2023-03-06
A lot of spaces in IOS.


giannis21:
#### Issue Description

<img width="347" alt="image" src="https://user-images.githubusercontent.com/48606335/223135957-3cca81a5-21c6-4ea6-abeb-a48cef25050d.png">

There is a lot of space in the top of messages in ios only. I use SafeAreaview. Also in the bottom i hide the bottom tab and i have put bottomOffset =(safeAreaInsets.bottom) ..I cannot remove these spaces..

#### Additional Information

* Nodejs version: v18.9.0
* React version: 17.0.2
* React Native version: 0.68.0
* react-native-gifted-chat version: 1.1.1
* Platform(s) (iOS, Android, or both?): IOS
* TypeScript version: 4.8.3

amerikan:
https://github.com/FaridSafi/react-native-gifted-chat/issues/2220#issuecomment-1153338844

Johan-dutoit:
Should be fixed with https://github.com/FaridSafi/react-native-gifted-chat/pull/2353 and https://github.com/FaridSafi/react-native-gifted-chat/releases/tag/v2.0.0

Issue #2349, opened 2023-02-27
Giftedchat react-native-safe-area-context breaks expo SDK 48


baltagih2:
#### Issue Description

On expo SDK 48, react-native-safe-area-context is upgraded to version 4.5.0, while gifted chat uses 4.4.1 as a peer dependency which leads to the registration of two views:

`Invariant Violation: Tried to register two views with the same name RNCSafeAreaProvider, js engine: hermes`

#### Steps to Reproduce / Code Snippets

- Upgrade expo to SDK 48
- Use latest version of react-native-gifted-chat (currently 1.1.1)
- App will crash with the following error:
<img width="966" alt="Screenshot 2023-02-27 at 12 13 10 PM" src="https://user-images.githubusercontent.com/62379853/221535984-a07b6dc2-923a-45e3-aa6e-a2189ad5dc14.png">

Running `npm ls react-native-safe-area-context` gives the following:

<img width="408" alt="Screenshot 2023-02-27 at 12 14 07 PM" src="https://user-images.githubusercontent.com/62379853/221536187-9aecd9b9-06c9-46c9-9998-48c0d7b30c14.png">

Running `npm dedupe` does not help with the issue since `"react-native-safe-area-context": "4.4.1"` is a dependency and not a peer dependency of react-native-gifted-chat 

#### Expected Results

- App should not crash
- GiftedChat should not force a specific version of the dependency

#### Additional Information

* Nodejs version: 16.15.0
* React version: 18.2.0
* React Native version: 0.71.3
* react-native-gifted-chat version: 1.1.1
* Platform(s) (iOS, Android, or both?): tested on iOS
* TypeScript version: 4.9.4

hramos:
Same here. I'm working around this, for the moment, with an override in my `package.json`:

```
  "overrides": {
    "react-native-safe-area-context": "4.5.0"
  },
```

baltagih2:
@hramos Thank you for your answer. I have tried that and also tried:

```
"overrides": {
    "react-native-gifted-chat": {
      "react-native-safe-area-context": "4.5.0"
    }
  },
```

but it doesn't fix the issue either

Tchaikovsky1114:
On expo SDK 48, i got same issue.
overrides isn't fix my issue..

hramos:
@baltagih2 what does `npm list react-native-safe-area-context` show? Here's my output, as you can see there's a single 4.5.0 version in use.

```
[...]

Issue #2319, opened 2022-12-22
Why doesnt my footer touch the end


shop-fluencer:
#### Issue Description

[FILL THIS OUT]

```
  return (
    <GiftedChat
      messages={messages}
      onSend={messages => onSend(messages)}
      onPressAvatar={handleNavigateToUser}
      user={{
        _id: 1,
      }}
    />
  )

```
[FILL THIS OUT]

Why my footer is not touching the end on iOS ? Look at image bottom

[FILL THIS OUT]

#### Additional Information

* Nodejs version: 16
* React version: expo 47
* React Native version: 0.70
* react-native-gifted-chat version: 1.1.0
* Platform(s) (iOS, Android, or both?): iOS
* TypeScript version: xxx

![WhatsApp Image 2022-12-22 at 15 51 35](https://user-images.githubusercontent.com/101994056/209161057-eb97e8c8-d4a0-4ad9-993a-98151047c6c7.jpeg)

fukemy:
try wrapInSafeArea={false}

visaals:
> try wrapInSafeArea={false}

That works in the simulator on a iPhone SE (thanks!), but not on an iPhone 12 still. Know any other workarounds we could try? Is it to adjust the `bottomOffset={22}` based on the device?

fukemy:
better use like this
const insets = useSafeAreaInset()

bottomInset= insets.bottom

Johan-dutoit:
Should be fixed with https://github.com/FaridSafi/react-native-gifted-chat/pull/2353 and https://github.com/FaridSafi/react-native-gifted-chat/releases/tag/v2.0.0

PR #2278, opened 2022-09-02
[fix] issue of Invariant Violation: Tried to register two views with the same name RNCSafeAreaProvider with expo SDK 46

kyoto21:

PR #2353, opened 2023-03-07
ref!: remove react-native-safe-area-context dependency

amerikan:
This is a proposal PR to remove the need to use `react-native-safe-area-context` as a dependency. From experience, it seems most developers handle `<SafeAreaView>`'s on their own terms. 

Furthermore, based on what I see, the GiftedChat library is currently using SafeAreaView so that the Example app can have spacing, but beyond that it doesn't seem to offer much more benefit or is a necessary thing.

In addition, a good chunk of issues are caused by GiftedChat trying to apply SafeAreaView without this being documented. Though I now see a PR by someone else for that https://github.com/FaridSafi/react-native-gifted-chat/pull/2336

Because the SafeAreaView dep is not being updated it often causes problems such as:

- https://github.com/FaridSafi/react-native-gifted-chat/issues/2349
- https://github.com/FaridSafi/react-native-gifted-chat/issues/2319
- https://github.com/FaridSafi/react-native-gifted-chat/issues/2352
- https://github.com/FaridSafi/react-native-gifted-chat/issues/2267
- https://github.com/FaridSafi/react-native-gifted-chat/issues/2251

And these PR's attempt to update the version 

- https://github.com/FaridSafi/react-native-gifted-chat/pull/2278
- https://github.com/FaridSafi/react-native-gifted-chat/pull/2286
- https://github.com/FaridSafi/react-native-gifted-chat/pull/2348

However, I still think it's better to just remove SafeAreaView from `<GiftedChat />` altogether. I feel like it will save a lot of headaches for the maintainers from having to always be updating `react-native-safe-area-context` when it practically adds no useful functionality to the library. I'm open to feedback. :)

### DIFF
--- a/example/App.tsx
+++ b/example/App.tsx
@@ -1,5 +1,6 @@
 import MaterialIcons from '@expo/vector-icons/MaterialIcons'
 import * as Linking from 'expo-linking'
+import { SafeAreaView } from 'react-native-safe-area-context'
 import AppLoading from 'expo-app-loading'
 import React, { Component } from 'react'
 import { StyleSheet, View, Text, Platform, Alert } from 'react-native'
@@ -20,7 +21,8 @@ import earlierMessages from './example-expo/data/earlierMessages'
 import { NavBar } from './components/navbar'
 
 const styles = StyleSheet.create({
-  container: { flex: 1 },
+  container: { flex: 1, backgroundColor: '#f5f5f5', },
+  content: { backgroundColor: "#ffffff", flex: 1, }
 })
 
 const filterBotMessages = message =>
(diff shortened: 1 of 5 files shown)


JotaMiller:
yes, please! I think that this library is incredible but I can't use it yet :(

amerikan:
@Johan-dutoit let me know what you think 😄

Johan-dutoit:
@amerikan I do agree with what you've said.  I'll have a proper review over the weekend perhaps and we'll take it from there.

amerikan:
@Johan-dutoit awesome, thank you! 😄

Johan-dutoit:
Code changes look pretty straight forward.  I haven't yet run the code, but hoping to do so soon.  Will release as a breaking change.

amerikan:
@Johan-dutoit fantastic! I just updated to 2.0.0 on my local project and no issues! Thanks for merging 😄 

Here are some other related PRs/Issues that you can close:

- https://github.com/FaridSafi/react-native-gifted-chat/pull/2336
- https://github.com/FaridSafi/react-native-gifted-chat/issues/2356
- https://github.com/FaridSafi/react-native-gifted-chat/issues/2343
- https://github.com/FaridSafi/react-native-gifted-chat/issues/2307
- https://github.com/FaridSafi/react-native-gifted-chat/issues/2220

Johan-dutoit:
Absolute legend @amerikan!, thanks for commenting and linking those issues/PRs.

I've closed them all.

joecampo:
@amerikan @Johan-dutoit Thanks to you both for getting rid of this dependency, will definitely help in maintenance!

One thing I've run into is that this PR breaks the `bottomOffset` prop from doing anything from what I can tell.

For example, if you're using React Navigation with a tab navigation bar, you'd typically do something like this to add some offset when the keyboard is shown (because of the tabs):

```js
bottomOffset={Platform.OS === 'android' ? undefined : tabBarHeight}
```

Prior to this PR, regardless if you were wrapping gifted in a safe view or not, the bottom offset would be applied when the keyboard was shown (and fix the gap between an input and keyboard when using tabs):

https://github.com/FaridSafi/react-native-gifted-chat/blob/75e75f0a7916379e7b4a3f3c25e8360f1e856d17/src/GiftedChat.tsx#L638

The `safeAreaSupport` had no concern to whether the view was a safe view or regular view, and would always be applied.

https://github.com/FaridSafi/react-native-gifted-chat/blob/75e75f0a7916379e7b4a3f3c25e8360f1e856d17/src/GiftedChat.tsx#L597-L599

Now that this line is removed, we can't dynamically account for the offset created by something like a React Navigation tab bar. I don't see setBottomOffset being applied anywhere else other than when the keyboard is hidden (where it properly sets it back to 0)?

Most folks using React Navigation tabs would use this combination of setting `wrapInSafe={false}` and setting the `bottomOffset` prop.

https://github.com/FaridSafi/react-native-gifted-chat/blob/d1b17f85ea32a115e43dcbbc8d3658483536bba5/src/GiftedChat.tsx#L642

amerikan:
@joecampo i see this removed that.

```diff
- this.setBottomOffset(this.safeAreaSupport(this.props.bottomOffset)) 
```

I think that fix would to just bring back the line, without the this.safeAreaSupport check? i.e.

```diff
+ this.setBottomOffset(this.props.bottomOffset) 
```

joecampo:
Thanks! I'll send a PR.

PR #2358, opened 2023-03-16
Fixes crash with: Tried to get frame for out of range index -1 [scrollToBottom (scrollToEnd of list with inverted option) only if messages available/loaded]

fahad86:
Without this fix we end-up having 
`Tried to get frame for out of range index -1` error thrown from the underlying FlatList

This is currently causing a lot of crashes on Production for us. The crashes happen when we try to navigate to the chat window directly from push notifications

### DIFF
--- a/src/GiftedChat.tsx
+++ b/src/GiftedChat.tsx
@@ -436,6 +436,9 @@ function GiftedChat<TMessage extends IMessage = IMessage>(
   const scrollToBottom = (animated = true) => {
     if (messageContainerRef?.current) {
       if (!inverted) {
+        if (!messages?.length) {
+          return;
+        }
         messageContainerRef.current.scrollToEnd({ animated })
       } else {
         messageContainerRef.current.scrollToOffset({


Johan-dutoit:
This seems like a sensible change.

manuhook:
@fahad86 maybe i'm missing something but i have this error when i'm trying to build : 
`src/GiftedChat.tsx:666:25 - error TS2339: Property 'data' does not exist on type 'Readonly<GiftedChatProps<TMessage>> & Readonly<{ children?: ReactNode; }>'.`

fahad86:
> @fahad86 maybe i'm missing something but i have this error when i'm trying to build : `src/GiftedChat.tsx:666:25 - error TS2339: Property 'data' does not exist on type 'Readonly<GiftedChatProps<TMessage>> & Readonly<{ children?: ReactNode; }>'.`

I guess the problem you mentioned would happen when there are no properties added to the <GiftedChat> tag?

Let me add some null checks.

src/GiftedChat.tsx
```diff
@@ -663,8 +663,11 @@ class GiftedChat<TMessage extends IMessage = IMessage> extends React.Component<
 
   scrollToBottom(animated = true) {
     if (this._messageContainerRef && this._messageContainerRef.current) {
-      const { inverted } = this.props
+      const { inverted, data } = this.props
```
fahad86, line 666:
```suggestion
      const inverted = this.props?.inverted;
      const data = this.props?.data;
```

Johan-dutoit:
👋 @fahad86  please resolve the conflicts, will merge in after.

fahad86:
> 👋 @fahad86 please resolve the conflicts, will merge in after.

Done.

src/GiftedChat.tsx
```diff
@@ -436,7 +436,10 @@ function GiftedChat<TMessage extends IMessage = IMessage>(
   const scrollToBottom = (animated = true) => {
     if (messageContainerRef?.current) {
       if (!inverted) {
-        messageContainerRef.current.scrollToEnd({ animated })
+        if (!messages.length) {
+          return;
+        }
+        _messageContainerRef.current.scrollToEnd({ animated })
```
fahad86, line 442:
```suggestion
        messageContainerRef.current.scrollToEnd({ animated })
```

fahad86:
@Johan-dutoit saw that there was one more conflict which I've resolved now

manuhook:
Hi @fahad86 ! we have also this issue, it will not be merged ?

fahad86:
> Hi @fahad86 ! we have also this issue, it will not be merged ?

I think with the refactoring for the latest version, empty messages handling has already been implemented in a different place. This PR should however still be valid for versions < 2.1.0

manuhook:
Problem was linked to react-native. Fixed in RN 0.72
https://github.com/facebook/react-native/issues/36850

PR #2408, opened 2023-06-15
updated link to ethora

Egatlov:
Added Ethora to the sponsor's list in the README file and updated the hyperlink


fukemy:
why bit.ly??

PR #2412, opened 2023-06-21
Expose quickReplyContainerStyle

robgraeber:
This adds a prop for adjusting the quickReplyContainer. This is useful for right aligning the quick replies and removing the max-width.


Johan-dutoit:
Thanks for the contribution!
I'll need to test this with the example app before merging, but seems like a good idea!

kesha-antonov:
Fixed in master

PR #2426, opened 2023-07-15
chore: fix to use strict equality

penicillin0:
Thank you for everything.

To prevent unexpected behavior due to implicit type conversion, I have made use of the strict equality operator.

This is my first time contributing to this repository, so please let me know if there are any oddities.


Johan-dutoit:
Hi @penicillin0, thanks for the contribution and welcome to the community!

The convenient thing about not using strict equality checks, is that it checks both null and undefined in one go.

Personally, I prefer strict checks, however with this change we need to consider the different flows here.  e.g. what will happen if we a null message is sent.

penicillin0:
thanks

PR #2432, opened 2023-08-07
fix import uuid v4 and upgrade module

preynal:
this was causing issues where the old uuid module didn't have the default export, and was conflicting with the more recent version. 
I don't think there is much other impact from this change.


rosariodev:
Why don´t merge this?

kesha-antonov:

Issue #2411, opened 2023-06-19
Messages disappear on second render on web


fkiDev:
#### Issue Description

When I enter to the gifted-chat screen, everything works fine on web and ios simulator, but when I go back and re-enter again the messages disappeared on the web but work on ios simulator

* React version: "18.1.0"
* React Native version: "0.70.8"
* react-native-gifted-chat version: "2.4.0"
* Platform(s) (iOS, Android, or both?): "web"
* Expo : "47.0.0"
* Expo : dev Client

axer4:
@fkiDev 
same issue, did u fix it?

quocvuong2000:
facing the same

edwinvrgs:
Same issue here with the following dependencies:
```json
    "expo": "~49.0.18",
    "react": "18.2.0",
    "react-dom": "18.2.0",
    "react-native": "0.72.6",
    "react-native-web": "~0.19.6",
    "react-native-gifted-chat": "^2.4.0",
```

PR #2437, opened 2023-08-20
fix: use messages from props instead of state

preynal:
it's unclear why the messages were attached to state, but there was some effect resetting them, so the chat view was empty, even if it had messages.

the message container now uses the messages from props.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "@types/react": "~16.9.35",
     "@types/react-native": "~0.63.2",
     "@types/react-test-renderer": "16.9.2",
-    "@types/uuid": "3.4.9",
+    "@types/uuid": "9.0.2",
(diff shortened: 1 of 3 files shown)


gajjardarshithasmukhbhai (approved):
LGTM!!

anwersolangi (changes requested):
LGTM!
I was facing the same issue on React Native Web on Expo, where, messages from props are appearing in console.logs, however, the state was not being updated. So, I think that this approach is good, however, if we use useCallback hook in the renderMessages function and pass messages as dependency array would make sure the re-render if any change happens in messages props.

edwinvrgs:
This seems to solve https://github.com/FaridSafi/react-native-gifted-chat/issues/2411.

deemyboy:
why is this not merged yet?

kesha-antonov:

PR #2471, opened 2024-01-17
Add initialNumToRender to the Props

changwoolab:
Hi! I added the `initialNumToRender` props for customizing the number of message bubbles at the initial rendering.


changwoolab:
Found I can use `listViewProps` for same purpose 😂
So I close this PR

PR #2474, opened 2024-02-01
fix: Remove potential race conditions from setState changes

brianoqr:
Hi, so i was using react-native-gifted-chat and had a race condition because my stream of messages happened at the same time as the "onLayout" event which made the isInitialized variable remain false as it was overwritten by a stale state value when the messages array was updated at the same time as onLayout.  This happens because setState is being called using a the local state value.  

I updated all the setState functions to use the state value passed by the function which is always up to date so you don't have any possibility of writing to state with a stale value.


NiharR27 (approved):

kesha-antonov:

PR #2533, opened 2024-09-09
Show typing indicator on web platform

cederache:
I don't understand why this condition is existing.
There is no way to show the typing indicator on the web platform if this condition is here.

I suggest to remove this condition to allow developer to show this indicator


kesha-antonov:
Hi
Thanks for suggestion
I think it's because previously we relied on 3rd party lib and now typing indicator rendered with reanimated
Did you try to launch example on web and see if typing indicator is working there?

cederache:
Yes, I tried it and it's working great

Issue #2523, opened 2024-08-29
Input gets cleared on keyboard dismiss


steavenb:
#### Input gets cleared on keyboard dismiss

While typing a message, if we click on the messages section and dismiss the keyboard, the input message is cleared

#### Expected Results

It should not clear, or there should be a prop that controls it.

ImBeCiliC:
This is due to a PR that added a reset of the text. This also happens when you switch to emoji keyboard.

This is the part that adds the issue:
https://github.com/FaridSafi/react-native-gifted-chat/pull/2493/files#diff-b11e3f8b30e19689a33f545b3567228eb41fd9cfbbad474b1b21e7b63eaa2d57R542-R545

@steavenb we fixed it with a custom composer and adding the prop text as null
`renderComposer={(composerProps: ComposerProps) => <ChatComposer {...composerProps} text={null} />}`

jorgtz:
Hello, is this going to be merged? It is an important bug!

pranshuchittora:
Facing the same issue :(

Issue #2504, opened 2024-06-20
Example build fails to run


vassantn:
#### Issue Description

I'm encountering an error when trying to run the example build

```
/home/user9390/Downloads/react-native-gifted-chat/node_modules/metro/node_modules/jest-haste-map/build/index.js:569
          throw new Error(message);
                ^

Error: jest-haste-map: Haste module naming collision:
  Duplicate module name: react-animated
  Paths: /home/user9390/Downloads/react-native-gifted-chat/node_modules/react-native/Libraries/Animated/release/package.json collides with /home/user9390/Downloads/react-native-gifted-chat/node_modules/expo/node_modules/react-native/Libraries/Animated/release/package.json

This error is caused by `hasteImpl` returning the same name for different files.
    at setModule (/home/user9390/Downloads/react-native-gifted-chat/node_modules/metro/node_modules/jest-haste-map/build/index.js:569:17)
    at workerReply (/home/user9390/Downloads/react-native-gifted-chat/node_modules/metro/node_modules/jest-haste-map/build/index.js:641:9)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Promise.all (index 21756)

Node.js v22.1.0
```

#### Steps to Reproduce / Code Snippets

1. git clone https://github.com/FaridSafi/react-native-gifted-chat
2. cd react-native-gifted-chat
3. npm install
4. expo start

#### Additional Information

* Nodejs version: Node.js version: v22.1.0
* npm version: 8.19.4

xuelink:
facing the same issue

```bash
Logs for your project will appear below. Press Ctrl+C to exit.
Failed to construct transformer:  Error: error:0308010C:digital envelope routines::unsupported
    at new Hash (node:internal/crypto/hash:68:19)
    at Object.createHash (node:crypto:138:10)
    at stableHash (/Users/xl/Developer/react-native-gifted-chat/example/node_modules/metro-cache/src/stableHash.js:19:8)
    at Object.getCacheKey (/Users/xl/Developer/react-native-gifted-chat/example/node_modules/metro-transform-worker/src/index.js:647:7)
    at getTransformCacheKey (/Users/xl/Developer/react-native-gifted-chat/example/node_modules/metro/src/DeltaBundler/getTransformCacheKey.js:32:19)
    at new Transformer (/Users/xl/Developer/react-native-gifted-chat/example/node_modules/metro/src/DeltaBundler/Transformer.js:180:28)
[...]

PR #2541, opened 2024-09-20
add isKeyboardInternallyHandled to allow custom logic again

ImBeCiliC:


ImBeCiliC:
@kesha-antonov this should resolve these issues + adding back the functionality of having a custom keyboard handler
Fixes:
https://github.com/FaridSafi/react-native-gifted-chat/issues/2523
https://github.com/FaridSafi/react-native-gifted-chat/issues/2504

JoniVR:
I agree that this option should be present, however, this current implementation doesn't remove the functionality entirely. The `useAnimatedKeyboard` reanimated hook also manipulates default adjustResize behavior. The hook is not unmounted in this implementation, so I assume this manipulation will remain present. See: https://docs.swmansion.com/react-native-reanimated/docs/device/useAnimatedKeyboard/#remarks

pranshuchittora:
This is a blocking issue, pls merge this.

Issue #2560, opened 2024-11-30
onPressActionButton is not anymore passed to `renderActions` or `Actions` component


Ksm125:
#### Issue Description

the props `onPressActionButton` is not passed to the callback `renderActions` and the `Actions` component, makes it impossible to have action on this now.

#### Steps to Reproduce / Code Snippets

```jsx
const handlePressAction = () => console.log('action pressed')

<GiftChat 
messages={[]}
onPressActionButton={handlePressAction}
/>
```

#### Expected Results

it should log the message in console, but nothing happens

#### Additional Information

* Nodejs version: 20.11.1
* React version: 18.2.0
* React Native version: 0.72.17
* react-native-gifted-chat version:  2.6.4
* Platform(s) (iOS, Android, or both?): both
* TypeScript version: 5.2.2

arnekolja:
Is this actually relevant? Seems only redundant to me, pass-drilling this while we also can just set it directly like

```
import { Send } from 'react-native-gifted-chat';

<GiftedChat
    renderActions={<Send onPressActionButton={handleActionButton} {...props} />}
/>
```

PR #2561, opened 2024-11-30
fix onPressActionButton not being called

Ksm125:
This add the `onPressActionButton` as props to `Actions` elements

Fix https://github.com/FaridSafi/react-native-gifted-chat/issues/2560

### DIFF
--- a/src/InputToolbar.tsx
+++ b/src/InputToolbar.tsx
@@ -42,6 +42,7 @@ export function InputToolbar<TMessage extends IMessage = IMessage> (
 
   const actionsFragment = useMemo(() => {
     const props = {
+      onPressActionButton,
       options,
       optionTintColor,
       icon,


kesha-antonov:
Hi

Doesn't understand why it's need since we already use this prop

https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/InputToolbar.tsx#L30

https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/InputToolbar.tsx#L55

Ksm125:
> Hi
> 
> Doesn't understand why it's need since we already use this prop
> 
> https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/InputToolbar.tsx#L30
> 
> https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/InputToolbar.tsx#L55

Hello @kesha-antonov , 

Well in fact this prop exists but it is not passed (at least not anymore) to the `renderActions` or `Actions`.  
When we create the `props` we do not have the `onPressActionButton` value in that object whereas in the previous library version, I have access to it within my component `renderActions`. Or is it what we expect now? 

Screenshot 2025-02-25 at 11 07 22



https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/InputToolbar.tsx#L51

kesha-antonov:
Understood what you meant
Thanks for the PR

Issue #2564, opened 2024-12-07
Type Updates and Expose onScroll event handler


NoodleOfDeath:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `react-native-gifted-chat@2.6.4` for the project I'm working on.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/react-native-gifted-chat/lib/GiftedChat.d.ts b/node_modules/react-native-gifted-chat/lib/GiftedChat.d.ts
index eb9b7de..a2a2cea 100644
--- a/node_modules/react-native-gifted-chat/lib/GiftedChat.d.ts
+++ b/node_modules/react-native-gifted-chat/lib/GiftedChat.d.ts
@@ -11,7 +11,7 @@ import { GiftedAvatar } from './GiftedAvatar';
 import { InputToolbar, InputToolbarProps } from './InputToolbar';
 import { LoadEarlier, LoadEarlierProps } from './LoadEarlier';
 import Message from './Message';
-import MessageContainer from './MessageContainer';
+import MessageContainer, { MessageContainerProps } from './MessageContainer';
 import { MessageImage, MessageImageProps } from './MessageImage';
 import { MessageText, MessageTextProps } from './MessageText';
 import { IMessage, LeftRightStyle, MessageAudioProps, MessageVideoProps, Reply, User } from './Models';
@@ -20,7 +20,7 @@ import { Send, SendProps } from './Send';
 import { SystemMessage, SystemMessageProps } from './SystemMessage';
 import { Time, TimeProps } from './Time';
 import * as utils from './utils';
-export interface GiftedChatProps<TMessage extends IMessage = IMessage> {
+export interface GiftedChatProps<TMessage extends IMessage = IMessage> extends Partial<MessageContainerProps<TMessage>> {
     messageContainerRef?: React.RefObject<FlatList<IMessage>>;
     textInputRef?: React.RefObject<TextInput>;
     messages?: TMessage[];
diff --git a/node_modules/react-native-gifted-chat/lib/MessageContainer.d.ts b/node_modules/react-native-gifted-chat/lib/MessageContainer.d.ts
index ded47b9..3bed4e3 100644
--- a/node_modules/react-native-gifted-chat/lib/MessageContainer.d.ts
+++ b/node_modules/react-native-gifted-chat/lib/MessageContainer.d.ts
[...]

PR #2565, opened 2024-12-07
expose renderTypingIndicator and handleOnScroll which are already pas…

NoodleOfDeath:
address the need for [this patch](https://github.com/FaridSafi/react-native-gifted-chat/issues/2564)

These properties are already passed to the underlying `MessageContainer` component, so might as well expose them to the developer

### DIFF
--- a/src/GiftedChat.tsx
+++ b/src/GiftedChat.tsx
@@ -66,7 +66,7 @@ import { useSafeAreaInsets } from 'react-native-safe-area-context'
 
 dayjs.extend(localizedFormat)
 
-export interface GiftedChatProps<TMessage extends IMessage = IMessage> {
+export interface GiftedChatProps<TMessage extends IMessage = IMessage> extends Partial<Omit<MessageContainer<TMessage>, 'scrollToBottom'>> {
   /* Message container ref */
   messageContainerRef?: React.RefObject<FlatList<IMessage>>
   /* text input ref */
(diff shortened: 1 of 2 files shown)


src/GiftedChat.tsx
```diff
@@ -66,7 +66,7 @@ import { useSafeAreaInsets } from 'react-native-safe-area-context'
 
 dayjs.extend(localizedFormat)
 
-export interface GiftedChatProps<TMessage extends IMessage = IMessage> {
+export interface GiftedChatProps<TMessage extends IMessage = IMessage> extends Partial<Omit<MessageContainer<TMessage>, 'scrollToBottom'>> {
```
NoodleOfDeath, line 69: had to omit `scrollToBottom` because of the name collision.. don't want to add a breaking change by renaming this

src/MessageContainer.tsx
```diff
@@ -168,7 +172,7 @@ export default class MessageContainer<
 
       return <LoadEarlier {...loadEarlierProps} />
     }
-    return null
+    return this.props.renderLoadEarlier?.(props)
```
NoodleOfDeath, line 175: exposes `renderLoadEarlier` to devs

src/MessageContainer.tsx
```diff
@@ -148,6 +150,8 @@ export default class MessageContainer<
   }
 
   renderTypingIndicator = () => {
+    if (this.props.renderTypingIndicator)
+      return this.props.renderTypingIndicator()
```
NoodleOfDeath, line 154: exposes `renderTypingIndicator` to devs

PR #2583, opened 2025-02-08
package.json: move ...get-random-values to dependencies

### LINKED ISSUE #2577: Missing dependency
Hi, `react-native-get-random-values` is defined in `devDependencies`, but is required at runtime, which causes an app using Gifted Chat to crash with the error:

```
Android Bundling failed 928ms (node_modules/expo-router/entry.js)
Unable to resolve "react-native-get-random-values" from "node_modules/react-native-gifted-chat/lib/index.js"
```

Please move the dependency to `dependencies` and release a fix.

adanielvv:
Fixes #2577

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "react-native-iphone-x-helper": "^1.3.1",
     "react-native-lightbox-v2": "^0.9.2",
     "react-native-parsed-text": "^0.0.22",
+    "react-native-get-random-values": "^1.11.0",
     "uuid": "^11.0.4"
   },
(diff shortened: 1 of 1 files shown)


kesha-antonov:
Hi!
Should be fixed by https://github.com/FaridSafi/react-native-gifted-chat/pull/2585

kesha-antonov:
It won't fix the issue since lib with native code should be installed on project-level

kesha-antonov:
More info here: https://github.com/react-native-community/cli/issues/870

PR #2585, opened 2025-02-11
Floating dates: make telegram like days animation & other refactoring for performance improvement and easier maintaining

kesha-antonov:
- Added smooth fade in/out of date while scrolling
- Made date appear on scroll to date between messages
- Added FlashList instead of FlatList
- Updated all deps
- Refactored few class components to functional components
- Refactored style usage to use common styles from 1 file
- Removed flow & prop-types since we have typescript and prop-types is deprecated and flow is not widely used

https://github.com/user-attachments/assets/a8bbcb6b-439a-4588-82fb-282968e6a30b


Randall71:
Hi @kesha-antonov , thanks for keeping this library up to date.

I saw that you replaced the use of FlatList with FlashList, wouldn't it be interesting to allow to add a list as a property that will be used? So for example, we can switch according to expectations, especially since we now have [legend-list](https://github.com/LegendApp/legend-list)

What do you think about it?

kesha-antonov:
> Hi @kesha-antonov , thanks for keeping this library up to date.
> 
> I saw that you replaced the use of FlatList with FlashList, wouldn't it be interesting to allow to add a list as a property that will be used? So for example, we can switch according to expectations, especially since we now have [legend-list](https://github.com/LegendApp/legend-list)
> 
> What do you think about it?

Hi
Depends on community demands

iM-GeeKy:
@kesha-antonov Is there a way to add styles to the floating date that was introduced?

TowhidKashem:
> > Hi @kesha-antonov , thanks for keeping this library up to date.
> > I saw that you replaced the use of FlatList with FlashList, wouldn't it be interesting to allow to add a list as a property that will be used? So for example, we can switch according to expectations, especially since we now have [legend-list](https://github.com/LegendApp/legend-list)
> > What do you think about it?
> 
> Hi Depends on community demands

I think this is a good idea, I was wondering the same, exposing a prop is what a lot of libs do in this type of scenario and is the most flexible. Personally I started to notice some issues with gifted chat recently and came in here looking for what might have changed. My personal experience with trying to roll my own flashlist based chat implementation was to end up using flatlist at the end cause visualization lead to a huge number of issues. So I'm kind of surprised this was merged as a minor change rather than a major one.

Issue #2527, opened 2024-09-03
Unable to scroll to the top of the chat when the keyboard is open, and it doesn’t automatically scroll to the bottom when a new message is sent.


nishmitha03:
#### Issue Description

1. Unable to scroll to the top of the chat when the keyboard is open.
2. The chat doesn’t automatically scroll to the bottom when a new message is sent

#### Steps to Reproduce / Code Snippets
1) Type a message in the chat.
2) Open the keyboard.
3) Attempt to scroll to the top of the chat.

Below is my code 

```
import React, {useCallback, useEffect, useState} from 'react';
import {SafeAreaView} from 'react-native';
import {GiftedChat} from 'react-native-gifted-chat';

const ChatComponent = () => {
  const [messages, setMessages] = useState([]);

  useEffect(() => {
    setMessages([
      {
        _id: 1,
        text: 'Hello developer',
        createdAt: new Date(),
        user: {
          _id: 2,
          name: 'React Native',
          avatar: 'https://placeimg.com/140/140/any',
        },
      },
    ]);
  }, []);

  const onSend = useCallback((messages = []) => {
    setMessages(previousMessages =>
      GiftedChat.append(previousMessages, messages),
    );
  }, []);

  return (
    <SafeAreaView style={{flex: 1}}>
      <GiftedChat
        messages={messages}
        onSend={messages => onSend(messages)}
        user={{
          _id: 1,
        }}
      />
    </SafeAreaView>
  );
};

export default ChatComponent;
```

```
import React from 'react';
import {SafeAreaProvider} from 'react-native-safe-area-context';
import ChatComponent from './ChatComponent';

const App = () => {
  return (
    <SafeAreaProvider>
      <ChatComponent />
    </SafeAreaProvider>
  );
};

export default App;
```

#### Expected Results

1. The chat should remain scrollable to the top even when the keyboard is open.
2. The chat should automatically scroll to the bottom when a new message is sent.

#### Additional Information

* Nodejs version: v18.16.0
* React version: 18.2.0
* React Native version: 0.72.0
* react-native-gifted-chat version: 2.6.1
* Platform(s) (iOS, Android, or both?):  IOS

Recording for -> Unable to scroll to the top of the chat when the keyboard is open.

https://github.com/user-attachments/assets/ec23b608-307a-4561-bf1b-93f98cd0e0cd

Recording for -> The chat doesn’t automatically scroll to the bottom when a new message is sent.

[...]

PR #2616, opened 2025-04-10
Fix: adjust scroll area when keyboard is open

riku99:
## Summary

This PR fixes the issue where the message list could not be fully scrolled when the keyboard is open, as described in [#2527](https://github.com/FaridSafi/react-native-gifted-chat/issues/2527).

## Changes

- Wrap `MessageContainer` with `KeyboardProvider`
- Dynamically adjust `contentContainerStyle` by adding `paddingBottom` based on the keyboard height
- Update snapshot tests to reflect the intended layout changes

## Reason

Without adjusting the padding, users cannot fully scroll to the latest messages when the keyboard is visible, especially when multiple messages are displayed.

## Related Issue

- Related to [#2527](https://github.com/FaridSafi/react-native-gifted-chat/issues/2527)

## Notes

- Only affects layout behavior when the keyboard is open
- No breaking changes for existing behavior

---

Please let me know if any further adjustments are needed!

### DIFF
--- a/src/MessageContainer/index.tsx
+++ b/src/MessageContainer/index.tsx
@@ -13,6 +13,7 @@ import Animated, { runOnJS, useAnimatedScrollHandler, useAnimatedStyle, useShare
 import { ReanimatedScrollEvent } from 'react-native-reanimated/lib/typescript/hook/commonTypes'
 import DayAnimated from './components/DayAnimated'
 import Item from './components/Item'
+import { useReanimatedKeyboardAnimation, KeyboardProvider } from 'react-native-keyboard-controller'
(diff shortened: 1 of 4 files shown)


josh-feldman:
@riku99 this approach has a few issues:
1. We are already inside a KeyboardProvider context so we don't need another one
2. (And more importantly), you are referencing an animated value `keyboard.height.value` in the react thread. You would need to use something like `useAnimatedStyle` to safely reference this shared value. That being said, I'm not sure this approach solves the issue either. See my comment on #2527 for a solution that I found to work for my case

Issue #2634, opened 2025-06-06
How to localize Today floating text


DimaIvashchuk:

DimaIvashchuk:
#2635

PR #2635, opened 2025-06-06
feat: added today localized object for Day component

DimaIvashchuk:


DimaIvashchuk:
#2634 I believe this is a good thing to have so we do not need to have custom components for renderDay

GaeCuure:
@DimaIvashchuk why did you close this PR?

Issue #2657, opened 2025-09-01
How to implement reply chat.


rabadiya-vishal:
As I want to implement reply to message as mentioned in attached screenshot.

<img width="1024" height="1252" alt="Image" src="https://github.com/user-attachments/assets/d4e7c26a-d1cf-4972-a765-59524c4b69c7" />

DimaIvashchuk:
As I am aware there is no solution out if the box

I had to implement my custom logic for that

I can share some details how I did it if you want

henrikra:
@kesha-antonov Do we have feature for reply feature?

kesha-antonov:
> @kesha-antonov Do we have feature for reply feature?

Not sure
Didn't use it with this lib

Would be cool to have it in the examples though

DimaIvashchuk:
@kesha-antonov  currently this lib does not have reply feature so I had to implement it using some hacks

in message bubble there is a props like renderAudio video etc, since my chat does not have video functionality I used renderVideo prop to render rely message. 
So it would be nice to add new prop to bubble renderReplyMessage and add to Message type a reply message with the same definition as currentMessage like id, user, text.

this will allow us at least to have an ability to create custom reply functionality without using existed props not by design


I can share an example how it was implemented in my project

P.S. It would be nice to have also an emoji reactions!! I implemented this one feature as well so if you are interested I can share it as well


here is a demo

https://github.com/user-attachments/assets/56ad6cc8-3765-408b-8ef4-bf6eb79dc2ec

kesha-antonov:
@DimaIvashchuk looks awesome!

PRs and code examples always welcome, thanks!

henrikra:
@DimaIvashchuk Please make PR!

DimaIvashchuk:
Hey @kesha-antonov @henrikra !
I created PR for reply message example 
#2690 
Sorry for possible mess, eslint didn’t work out of the box and I didnt bother to fix it

tell me what do you think about it, you may use it as a ref to add this functionality to your lib

I can also create an example PRs that will show how to implement REACTIONS and MENTIONS  if needed
Unfortunately I do not have enough time at this moment to contribute to the lib functionality as well

DimaIvashchuk:
@kesha-antonov created a PR for reply message functionality #2692

cyrus25:
@DimaIvashchuk can you also share the implementation for reactions?

kesha-antonov:
> #2692

[...]

PR #2690, opened 2025-12-07
feat: added reply message implementation example

DimaIvashchuk:
This PR adds a minimal reply feature for Gifted Chat. Since the library doesn’t support replies natively, we use renderImage (or other render props) to display a reply preview. Message objects now include the required flags (image: '1', audio: '1', etc.) so Gifted Chat knows which renderer to use.


DimaIvashchuk:
#2657 related to this issue

kesha-antonov:
Thanks 🙏

Issue #2697, opened 2025-12-21
TypeScript errors in v3.2.2 with strict type checking


henrikra:
## Description

  When running TypeScript with `--noEmit` (strict type checking), the library produces 3 type errors that prevent projects from compiling cleanly.

  ## Environment

  - react-native-gifted-chat version: 3.2.2
  - TypeScript version: 5.9.2
  - React Native version: 0.81.5

  ## Errors

  ### 1. Unused `@ts-expect-error` directive in `Composer.tsx`

  **File:** `src/Composer.tsx:52`
  **Error:** `TS2578: Unused '@ts-expect-error' directive`

  The `@ts-expect-error` comment is no longer needed as the underlying type issue was likely fixed.

  ### 2. Invalid `disabled` prop on `TouchableOpacity` in `GiftedAvatar.tsx`

  **File:** `src/GiftedAvatar.tsx:162`
  **Error:** `TS2322: Property 'disabled' does not exist on type`

  The `TouchableOpacity` from `react-native-gesture-handler` doesn't have a `disabled` prop in its type definitions.

  ### 3. Generic type mismatch in `SystemMessage.tsx`

  **File:** `src/SystemMessage.tsx:38`
  **Error:** `TS2322: Types of property 'onPress' are incompatible`

  The `MessageText` component receives props with a generic `TMessage` type, but the `onPress` handler expects `IMessage`, causing a type incompatibility when `TMessage` could be a different subtype.

  ## Steps to Reproduce

  1. Install `react-native-gifted-chat@3.2.2`
  2. Run `tsc --noEmit`

  ## Expected Behavior

  Library should compile without TypeScript errors.

henrikra:
@eisodev-r are you getting the same?

juliussneezer04:
+1 on this, until the PR is merged I've been using this patch for the strict `MessageText` typing bug:
```patch
diff --git a/src/MessageText.tsx b/src/MessageText.tsx
index d20d627bd791a19dd1f2ff280391f5ffee0c652c..9b5d6e0757fb3754c3d3a0ca40c659f7c18e5afe 100644
--- a/src/MessageText.tsx
+++ b/src/MessageText.tsx
@@ -35,7 +35,7 @@ export type MessageTextProps<TMessage extends IMessage> = {
   stripPrefix?: boolean
 }
 
-export const MessageText: React.FC<MessageTextProps<IMessage>> = ({
[...]

PR #2699, opened 2025-12-22
Fix package distribution by shipping compiled JavaScript instead of TypeScript source

henrikra:
The package is currently configured to distribute raw TypeScript source files instead of compiled JavaScript:

  - main points to src/index.ts (TypeScript source)
  - types points to src/index.ts (should be .d.ts declaration file)
  - files only includes src/ directory
  - prepublishOnly doesn't compile TypeScript before publishing

  This means the published npm package is broken for most consumers because:

  1. Plain JavaScript projects can't use it - They cannot import .ts files directly
  2. Most bundlers fail - Webpack, Rollup, and other bundlers don't transpile node_modules by default
  3. TypeScript projects need extra config - Users would need to configure their build tools to compile this dependency, which is non-standard

  Solution

  - Changed main to lib/index.js (compiled JavaScript)
  - Changed types to lib/index.d.ts (proper type declarations)
  - Changed files to include lib/ instead of src/
  - Added yarn tsc:write to prepublishOnly to ensure the package is compiled before each publish

  Impact

  Without this fix, anyone installing the package from npm would get unusable TypeScript source files. With this fix, the package correctly ships compiled JavaScript that works out of the box for all consumers, along with type declarations for TypeScript users.

Testing
You can test with `npm pack --dry-run` which files are now included in the package. Remember to run `yarn tsc:write` first!

PS I dropped Node 18 because CI failed (https://github.com/FaridSafi/react-native-gifted-chat/actions/runs/20430123554/job/58698813148?pr=2699) and Node 18 security support ended 7 months ago: https://endoflife.date/nodejs


henrikra:
@kesha-antonov another easy one! Can you you publish after this one merged? New patch?

henrikra:
@kesha-antonov now this is easy to merge also 👍

henrikra:
@kesha-antonov please tell me if you need clarification. TLDR this PR distributes compiled javascript code like libraries should instead of typescript source. This will prevent TS errors like this from the future as well: https://github.com/FaridSafi/react-native-gifted-chat/issues/2697

kesha-antonov:
Hi
I think it's good to include src too because many times in my dev experience I needed to look into source code and tweak things around. If it's compiled js only you need to install version from GitHub, make changes there

PR #2705, opened 2026-01-15
Fix TypeScript errors, add type check to CI, and drop Node 18

henrikra:
Description:
  - Fix tsc errors in MessagesContainer
  - Add type check step to PR CI
  - Drop deprecated Node 18 from CI matrix

  These changes fix existing TypeScript errors that had gone unnoticed. To prevent this from
  happening again, type checking is now part of the PR workflow. This helps catch type errors
  before they get merged and ensures contributors maintain type safety going forward. Additionally,
   Node 18 has been removed since it reached end-of-life and is no longer supported.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [18, 20]
+        node-version: [20, 22]
 
     steps:
       - uses: actions/checkout@v3


henrikra:
@kesha-antonov Easy PR to check. There was still TSC error in the latest version. Also added tsc to PRs so these tsc errors wont go unnoticed ever again

kesha-antonov:
Hi
Thanks for the PR!
There's conflicts but i'll add your changes

PR #2720, opened 2026-02-23
feat: implements messagesContainerInsets prop

kaarl77:
This prop is needed when the GiftedChat is rendered in an iOS 26 screen with Liquid Glass / transparent header. Currently, there was no way to set the content's inset, or starting point, without margin/padding, which wouldn't allow the messages to fully go behind the header

Before (Load earlier button not visible or interactable):

https://github.com/user-attachments/assets/a1838d7e-077b-4b07-9beb-9153455b297c

After (Load earlier button visible and interactable):

https://github.com/user-attachments/assets/0547ebb0-db6e-4188-94f2-a83070f8903b


kesha-antonov:
Hi

Thanks for the PR

You can pass this props through `listProps`: https://github.com/FaridSafi/react-native-gifted-chat#messages--message-container

PR #2722, opened 2026-02-23
chore: ship compiled `.js` and `.d.ts` instead of raw TypeScript source

### LINKED ISSUE #2712: Typescript errors in v3.3.2 with strict typechecking
My tests are failing because of typescript errors in this package.

Running `tsc` command with my tsconfig that has `"strict": true` and `"noUncheckedIndexedAccess": true` yields these errors:
```
Found 23 errors in 10 files.

Errors  Files
     1  node_modules/react-native-gifted-chat/src/Actions.tsx:59
     1  node_modules/react-native-gifted-chat/src/Bubble/index.tsx:376
     1  node_modules/react-native-gifted-chat/src/Composer.tsx:76
     3  node_modules/react-native-gifted-chat/src/GiftedAvatar.tsx:66
     1  node_modules/react-native-gifted-chat/src/Message/index.tsx:179
     6  node_modules/react-native-gifted-chat/src/MessagesContainer/components/DayAnimated/index.tsx:42
     3  node_modules/react-native-gifted-chat/src/MessagesContainer/components/Item/index.tsx:45
     3  node_modules/react-native-gifted-chat/src/MessagesContainer/index.tsx:176
     1  node_modules/react-native-gifted-chat/src/Time.tsx:56
     3  node_modules/react-native-zoom-reanimated/src/index.tsx:562
```

and running it with just `"strict": true` (removing `noUncheckedIndexedAccess`) yields:

```
Found 5 errors in 5 files.

Errors  Files
     1  node_modules/react-native-gifted-chat/src/Bubble/index.tsx:376
     1  node_modules/react-native-gifted-chat/src/Composer.tsx:76
     1  node_modules/react-native-gifted-chat/src/Message/index.tsx:179
     1  node_modules/react-native-gifted-chat/src/Time.tsx:56
     1  node_modules/react-native-zoom-reanimated/src/index.tsx:1060
```

It seems this is a regression since this [same issue](https://github.com/FaridSafi/react-native-gifted-chat/issues/2697) was created and apparently fixed by #2698. However, the latest version `v3.3.2` is still showing errors. 

I believe this is will be an easy fix of adding some `?`s on variables that could possibly be undefined and fixing some `TextStyle`s that should be `ViewStyle`s.

Thank you for your work 🙏🏼

Best,
Cash

gianlucadifrancesco:
First of all, thanks for the amazing library!

Since I had some TS errors and wanted to definitely fix them, here is a PR that focuses on making sure the project won't have any TS problems to consumers.

In order to do so, this PR will:

- **Ship compiled output instead of raw `.ts`/`.tsx` source files.** The library currently publishes raw TypeScript source via `"main": "src/index.ts"`, which causes consuming projects' `tsc` to type-check the library as source code — even with `skipLibCheck: true` (which only applies to `.d.ts` files).
This change compiles to `lib/` and publishes `.js` + `.d.ts` + `.d.ts.map` files instead.
This will also fix #2712 and prevents similar errors in the future.
- **Add proper build scripts** (`build`, `typecheck`) and wire `prepublishOnly` to run `lint + test + build` before every publish.
- **Export `Color` and `styles`** from the public API. They were previously internal, but I hit a use case where they were necessary and figured they’d be helpful for others as well.

⚠️ This might be a breaking change if someone imports things from `/src/...`, since with this PR everything will directly be exported from `react-native-gifted-chat` itself.

## Motivation

When a consuming project uses this library via npm/yarn, TypeScript resolves imports to raw `.ts` files inside `node_modules`. Since these are not `.d.ts` declaration files, `skipLibCheck: true` has no effect and `tsc` **fully type-checks** the library on every build.

This slows down compilation and can surface errors from the library's internals that are irrelevant to consumers.

image

## Notes

| File | Change |
|---|---|
| `tsconfig.json` | Added `declarationMap: true` for better "Go to Definition" DX in consuming projects. |
| `.github/workflows/main.yml` | CI type-check step now uses `yarn typecheck`. |


gfernandez-me:
Can someone review this PR?

cashc:
@kesha-antonov are you the maintainer of this library? I would really appreciate you reviewing and getting this in a new release to fix the typescript errors in my project 🙏. Thank you.

Titozzz:
@kesha-antonov Thanks for merging ! Can that make it into a release any time soon ?

cashc:
@kesha-antonov releasing a new version with this change would be much appreciated 🙏 thank you!

gzurbach:
@kesha-antonov Thanks again for merging this fix, it’s hugely appreciated 🙏

For a bit of context on why a release would help us so much: the current published version produces TypeScript errors during type checking in our app, so our CI has effectively been “red by default” for the past couple of months because of react-native-gifted-chat.

That means every PR requires manually inspecting the type errors to figure out whether they’re coming from the library or from newly introduced issues in our codebase. It’s pretty error-prone, and we’ve already had a few typing issues slip through because of the noise.

A new npm release with this fix would make a really big difference for our team and CI reliability. Thanks again for all the work maintaining the library.

kesha-antonov:
> @kesha-antonov Thanks again for merging this fix, it’s hugely appreciated 🙏
> 
> For a bit of context on why a release would help us so much: the current published version produces TypeScript errors during type checking in our app, so our CI has effectively been “red by default” for the past couple of months because of react-native-gifted-chat.
> 
> That means every PR requires manually inspecting the type errors to figure out whether they’re coming from the library or from newly introduced issues in our codebase. It’s pretty error-prone, and we’ve already had a few typing issues slip through because of the noise.
> 
> A new npm release with this fix would make a really big difference for our team and CI reliability. Thanks again for all the work maintaining the library.

Hi
Thanks for the explanation!
Published

gzurbach:
Thanks a lot @kesha-antonov 🙏

PR #2725, opened 2026-03-08
feat/message-reactions

DimaIvashchuk:
I have added a minimal working functionality for message reactions.
@kesha-antonov there might be a mess, sorry for that:) but I hope you will get basic understanding.
There are couple of things to consider:

1. Full modal with emojies comes from my personal npm lib that I have created for my self, it is not really well tested and there are couple of things that can be improved.
2. In the original implementation I use bottom sheet from gorham. here is not. You might consider using it
3. also the animations might be uncompleted or inappropriate.
4. Also i didn't add an ability to fast react on double tap, but it should be pretty much  straightforward


kesha-antonov:
Thanks @DimaIvashchuk! I rebased this onto current `master` and did a cleanup pass so it can land:

- **Rebased & conflicts resolved** against the latest `master` (the animated day header work had since touched `Bubble`).
- **Dropped the unrelated reformatting** so the diff is reactions-only (now ~1.3k insertions vs the original ~2k/600).
- **No new core dependencies.** The full emoji browser (`react-native-emoji-chooser` + `react-native-svg` + `async-storage`) moved entirely into the example, wired via the `renderReactionPicker` override. The library itself only uses `react-native-gesture-handler` and `react-native-reanimated`, which are already peer deps. Core ships a lightweight quick picker; consumers opt into a full browser if they want one.
- **Added a README section** (usage, props, screenshots) and an example screen.
- **Verified on an Android emulator**: quick picker, reaction pills (with counts + active state), and the full emoji browser all render correctly with no native crash.

Your original commit is preserved as the feature commit (authored by you); the docs are a follow-up commit. Merging once CI is green.

